PR #7, opened 2015-08-10
implemented LagrangeQuadratic order 2

YetAnotherMinion:
The doxygen indicates that getLagrange for dimensions 1 and 2 are implemented, while in fact dimension 2 is not.

```
Get the Lagrangian shape function of some polynomial order.
we have only first and second order so far
```

I think I got the shape functions right, but please check. My motivation is to get apf::Mesh::changeShape(apf::getLagrange(2)) to work and automatically create the 9 nodes. I do see where the new nodes get added to the mesh, so if you can point that file out I can also address that too. I could not find the ordering of the nodes in eta space, so I went with ordering in Hughes. If the shape functions and gradients are correct, then it should be trivial to reorder them by someone who does know the ordering.

### DIFF
--- a/apf/apfShape.cc
+++ b/apf/apfShape.cc
@@ -465,20 +465,71 @@ class LagrangeQuadratic : public QuadraticBase
     LagrangeQuadratic() { registerSelf(apf::LagrangeQuadratic::getName()); }
     const char* getName() const {return "Lagrange Quadratic";}
     class Quad : public EntityShape
-    { /* TODO: implement this and then update hasNodesIn, countNodesOn */
+    {
       public:
         void getValues(Mesh*, MeshEntity*,
-            Vector3 const&, NewArray<double>&) const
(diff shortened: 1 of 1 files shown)


YetAnotherMinion:
Also it makes the shape functions much easier to read when you substitute a single symbol for xi[0], xi[1] ,... It will get optimized out by compiler anyway and makes the tight grouping style of arithmetic operators favored by this code base a little easier to understand. If the shape functions look good one can always do a local find and replace to put those xi[i]'s back if that is the convention

ibaned:
yea, theres a whole funny story about the serendipity quad but the short version is I was lazy and you seem to have wrapped it up nicely so I'll pull this and run a couple sanity tests.

Thanks !

by the way, the nodes will take care of themselves, that countNodes function controls it all.
also, the single-char names are fine, but its dangerous to go back and change the existing
code so I'll just take this new stuff as is.

PR #10, opened 2015-08-12
Enabled editing of matricies after construction

YetAnotherMinion:
previously one could not edit the contents of an apf::Matrix after it had been constructed.
this was annoying when trying to allocate an array of apf::Matrix and having the default constructor called for every matrix. Now you can actually use the matrix in the array.

### DIFF
--- a/apf/apfMatrix.h
+++ b/apf/apfMatrix.h
@@ -40,6 +40,18 @@ class Matrix : public Array<Vector<N>,M>
       for (std::size_t j=0; j < N; ++j)
         this->elements[i][j] = array[i][j];
     }
+
+    /** \brief immutable index operator */
+    double operator()(std::size_t i, std::size_t j) const
+    {
+      return this->elements[i][j];
+    }
+    /** \brief mutable index operator */
+    double& operator()(std::size_t i, std::size_t j)
+    {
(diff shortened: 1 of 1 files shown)


bgranzow:
I think it's fine to add that type of access operator, but I'm not going to make the final call to merge the pull request. Your claim about not being able to edit entries of a matrix after construction isn't true, though. The code below shows how you can do that.

```
#include "apfMatrix.h"

int main()
{
  apf::Matrix<2,2> A;
  A[0][0] = 1.0; A[0][1] = 0.0;
  A[1][0] = 0.0; A[1][1] = 1.0;
  std::cout << A;
}
```

YetAnotherMinion:
I had tried that at the beginning, and got a segfault, which was what prompted trying to add the operators. Looking back right now, I got the segfault because I did this

```C++
apf::NewArray< apf::Matrix< 3,2 > > B;
/*construct each of the nnodes shape function matricies*/
uint32_t ii;
for(ii = 0; ii < this->nnodes; ++ii) {
    B[ii][0][0] = gradShape[ii][0];

```

I had forgotten to allocate the array! Which I only  discovered after I had made the changes and tested the new methods. 

I withdraw the pull request since it is not needed.

PR #24, opened 2016-07-11
can : make NewArray newer

bgranzow:
This commit is intended to:

* Remove the unnused swap method

* Endow the NewArray container with size information by
  inheriting from the dynamic can Array container. This
  should make calls to NewArray->allocate(n) faster for
  NewArrays that have already been allocated to size n.

If this commit breaks code you've written that I haven't
accounted for, let me know @bgranzow

### DIFF
--- a/can/canNewArray.h
+++ b/can/canNewArray.h
@@ -12,6 +12,7 @@
     \brief wrapper for operator new/delete [] */
 
 #include <cstddef>
+#include "canArray.h"
 
 namespace can {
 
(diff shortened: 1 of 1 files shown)


bgranzow:
I'm making this a pull request in case anyone wants to sanity check the little work that I did. The regression tests all passed with this commit.

ibaned:
The core-sim tests passed also. I'll go ahead with this.

PR #117, opened 2017-07-21
Develop

mortezah:
This is the shape fix stuff for curve adapt plus some other changes/bug fixes along the way.


apf/apfSimplexAngleCalcs.cc
```diff
@@ -0,0 +1,366 @@
...
+  Matrix3x3 J;
+  getJacobian(faceElem, xi, J);
+  destroyMeshElement(faceElem);
+
+  J = getJacobianInMetric(J, Q);
+
+  return cross(J[0], J[1]).normalize();
+}
+
+/* void computeEdgeEdgeTangentsInTri(Mesh* m, MeshEntity* tri, MeshEntity* e1, MeshEntity* e2, */
```
cwsmith: Please remove commented code.

apf/apfSimplexAngleCalcs.cc
```diff
@@ -0,0 +1,366 @@
...
+    case 1:
+      xi = Vector3(0.5, 0.5, 0.0);
+      break;
+    case 2:
+      xi = Vector3(0.0, 0.5, 0.0);
+      break;
+    default:
+      break;
+  }
+  /* Vector3 xi_2 = apf::boundaryToElementXi(m, edge, face, Vector3(0.0, 0.0, 0.0)); */
```
cwsmith: Please remove commented code.

apf/apfSimplexAngleCalcs.cc
```diff
@@ -0,0 +1,366 @@
...
+    t = Vector3(0.0, 0.0, 0.0) - Vector3(J[0]);
+  return t.normalize();
+}
+
+
+Vector3 computeFaceNormalAtVertex(Mesh* m, MeshEntity* face,
+    MeshEntity* vert,
+    const Matrix3x3& Q)
+{
+  PCU_ALWAYS_ASSERT(m->getType(face) == Mesh::TRIANGLE);
```
cwsmith: Is there some check in the higher level APIs that stops folks with mixed meshes from hitting this assert?

apf/apfSimplexAngleCalcs.cc
```diff
@@ -0,0 +1,366 @@
...
+  MeshEntity* fe[3];
+  m->getDownward(face, 1, fe);
+  if (findIn(fe, 3, edge) > -1)
+    return 1.0;
+
+
+  // find the share vertex
+  MeshEntity* ev[2];
+  MeshEntity* fv[3];
+  MeshEntity* sharedVert = 0;
```
cwsmith: I think an MDS vertex (MeshEntity*) with id 0 is possible...

crv/crvQuality.cc
```diff
@@ -638,7 +638,7 @@ int Quality3D::computeJacDetNodes(apf::MeshEntity* e,
       interNodes[index] = apf::getDV(me,xi[index]);
       if(interNodes[index] < 1e-10){
         apf::destroyMeshElement(me);
-        return i+8;
+        return edge+8;
```
cwsmith: What does this do? (same question for face+14 below)

crv/crvShape.cc
```diff
@@ -300,66 +304,65 @@ class EdgeReshaper : public ma::Operator
...
+static bool isCornerTriAngleLargeMetric(crv::Adapt *a,
     ma::Entity* tri, int index)
 {
   ma::Mesh* m = a->mesh;
-  apf::Element* elem = apf::createElement(m->getCoordinateField(),tri);
-  apf::NewArray<apf::Vector3> nodes;
-  apf::getVectorNodes(elem,nodes);
-  apf::destroyElement(elem);
+  // get downward vertexes
+  ma::Entity* down[3];
```
cwsmith: Maybe assert that `tri` is a triangle?

crv/crvShape.cc
```diff
@@ -300,66 +304,65 @@ class EdgeReshaper : public ma::Operator
...
-      ma::Vector xi = crv::elem_vert_xi[apf::Mesh::TET][tetIndex];
-      apf::getJacobian(me,xi,J);
-      apf::destroyMeshElement(me);
-
-      if(apf::getJacobianDeterminant(J,3) < a->input->validQuality){
-        return true;
-      }
-    }
-  }
+  // TODO: Take care of this later!
```
cwsmith: Please remove commented code.

crv/crvShape.cc
```diff
@@ -535,7 +521,10 @@ static int markEdgesToFix(Adapt* a, int flag)
 
   ma::Mesh* m = a->mesh;
   ma::Entity* e;
-  ma::Entity* edges[3];
+
+  // markEdges could have upto 6 edges marked!!!
+  /* ma::Entity* edges[3]; */
```
cwsmith: for tets.  Are there any plans to expand this code to support hexs, pyramids or prisms?

crv/crvShapeFixer.cc
```diff
@@ -0,0 +1,349 @@
...
+*/
+
+int getCrvSliverCode(
+    Adapt* a,
+    apf::MeshEntity* tet)
+{
+  ma::SizeField* sf = a->sizeField;
+  apf::Mesh2* m = a->mesh;
+  apf::Matrix3x3 J,Q;
+  apf::MeshElement* me = apf::createMeshElement(m,tet);
```
cwsmith: Assert that `tet` is a tet?

crv/crvShapeHandler.cc
```diff
@@ -265,28 +265,35 @@ class BezierHandler : public ma::ShapeHandler
...
-        double A = 0.5*apf::cross(p[1]-p[0],p[2]-p[0])[2];
-        double s=0;
-        for (int i=0; i < 3; ++i)
-          s += l[i]*l[i];
-        double lq;
-        if (A < 0)
-          lq = -48*(A*A)/(s*s);
-        else
-          lq = 48*(A*A)/(s*s);
+	/* The following commented version does not account for
```
cwsmith: please remove

ma/maSnapper.cc
```diff
@@ -130,49 +129,66 @@ static bool tryDiggingEdge(Adapt* adapter, Collapse& collapse, Entity* e)
...
 
-#ifdef DO_FPP
-  static bool trySnappingToFPP2(Adapt* a, Collapse& c, Tag* st, Entity* v,
-      apf::Up& badElements)
-  {
-    // make a FPPSnapper Object
-    FPPSnapper fpps(a, c, st, v, badElements);
-    return fpps.findFPP() ? fpps.snapToFPP() : false;
-  }
+/* #ifdef DO_FPP */
```
cwsmith: please remove

apf/apfSimplexAngleCalcs.cc
mortezah: @cwsmith  what did you mean by this?

cwsmith: The trick here is that the MeshEntity* is actually the id of MDS entity.  But, I was wrong and MDS starts ids at 1.  If they did start at 0 then the assert on line 292 could trip on a valid case.

crv/crvQuality.cc
cwsmith: Hmm. How about using an enum or a function to make this more readable?

crv/crvShape.cc
mortezah: I would like to guess that maybe at some point in the future. Also, I think (need to do a bit more thinking on this) it would be better to deal with non-simplex cases separately. The current curving and curve adapt features rely heavily on the assumptions that we are dealing with simplexes.

crv/crvShapeFixer.cc
mortezah: I don't think that is necessary. Only adapt developer may need to use this.

crv/crvShape.cc
mortezah: not necessary. This happens inside a static function which is only called from other functions the including .cc file. Something else would fail much earlier before this call.

crv/crvQuality.cc
mortezah: Downstream code (eg https://github.com/SCOREC/core/blob/master/crv/crvShape.cc#L55) relies on this numbering/marking scheme.

apf/apfSimplexAngleCalcs.cc
ibaned: Yea, there is a `+1` in the formula exactly for this reason, to get it to behave like a real pointer so a zero pointer becomes a `-1` index.

PR #130, opened 2017-11-07
Fix SCOREC configure for TriBITS build (TriBITSPub/TriBITS#200)

bartlettroscoe:
This fixes the SCOREC configure under TriBITS with Trilinos.  In this case, TriBITS was creating a terrible and misleading error message.  If fixed TriBITS to create a much better error message for this use case.  Hopefully this will save people time in the future for the same mistake.


ibaned:
Thanks @bartlettroscoe !

PR #144, opened 2018-03-19
add equality operator for apf::Matrix

timetravellers:
This functionality was needed for writing some unit tests.


cwsmith:
I'm not sure what context this is intended for, but for general use I would suggest looking into something like this for comparing floating point values:
https://github.com/google/googletest/blob/dccc2d67547a5a3a97e4f211f39df931c6fbd5d5/googletest/include/gtest/internal/gtest-internal.h#L334-L347
as suggested/discussed here:
https://stackoverflow.com/a/3423299

timetravellers:
Sure I can do something like that...Does core have a fuzz/tolerance constant that is defined for this sort of operation?

timetravellers:
@cwsmith  Does this look reasonable to you? I took a look at the google-test code that you suggested looking at and I went with something less complicated for now. The on google-test licensing is such that we might be able to just include their float comparison function.

cwsmith:
Can you modify the comment above the `isClose` function to point at a reference for your comparison procedure?

The original pull request was from your master branch to scorec master.  I switched the merge target to scorec develop; the nightly system will merge to master if the build is clean and all tests pass.  After doing this I see a boat load of merge commits.  Can you rebase your master branch against scorec/develop and see if that cleans that up?  This may not work and we'll have to recreate the PR...

PR #158, opened 2018-07-23
AggregateNumbering

tobinw:
Implementation of a read-only numbering type with minimal memory footprint to allow for novel block-structures in the matrices resulting from finite element assembly using these numbering. Mostly for use in the fusion projects, but they can be used as regular a Numbering as well if the user simply wants to reduce memory usage. Aggregation scopes are determined by MPI_Comms passed to the creation functions but the creation is still a globally parallel process across the typical PCU comm.

They inherit from the Numbering type and require only a singe integer or long per node, they are treated by the Mesh as normal Numbering or GlobalNumberings, and are written out to VTK as expected.

They cannot be written to so they cannot be synchronized (though they are synchronized during creation so are globally valid following creation).

Minimal changes have been made outside of the new files to introduce the new concept. Possibly the most impactful is changing several member functions of the NumberingOf class to be virtual member functions.


apf/apfAggregateNumbering.cc
```diff
@@ -0,0 +1,421 @@
...
+    {
+      PCU_Switch_Comm(pcu_cm);
+      delete init_share;
+    }
+    // during init the static sharing may be used which
+    //  only provides binary ownership information, not
+    //  the owner rank of unowned entities, so we replace
+    //  it with the passed-in sharing here before globalization
+    //  which requires the rank information
+    n->setSharing(share);
```
tobinw: This may not be required anymore since StaticSharing now supplies ownership information as well.. but may also be the cleaner solution in the end

apf/apfAggregateNumbering.h
```diff
@@ -0,0 +1,37 @@
...
+  class Field;
+  class FieldShape;
+  class Sharing;
+  class Mesh;
+  class MeshEntity;
+  template <class T>
+  class AggregateNumberingOf;
+  typedef AggregateNumberingOf<int> AggNumbering;
+  typedef AggregateNumberingOf<long> GlobalAggNumbering;
+  AggNumbering * createAggNumbering(Field * f,
```
tobinw: This should probably just return Numbering* since AggNumbering inherits from Numbering and the entire point of the class is to be used as a Numbering

apf/apfAggregateNumbering.h
```diff
@@ -0,0 +1,37 @@
...
+                                    Sharing * share = NULL);
+  AggNumbering * createAggNumbering(Mesh * m,
+                                    const char * name,
+                                    FieldShape * shape,
+                                    int blocks,
+                                    int dofs_per_block,
+                                    MPI_Comm cm,
+                                    Sharing * share = NULL);
+  void destroyAggNumbering(AggNumbering * n);
+  int getElementNumbers(AggNumbering * n, MeshEntity * e, NewArray<int> & numbers);
```
tobinw: Ah but this needs an AggNumbering specifically rather than just a Numbering?

tobinw: Oh I might be able to just get rid of this since the underlying FieldData is now a ```UserData<int/long>``` which should operated correctly in the ```FieldDataOf<T>::getElementData()``` function I think.

apf/apfNumbering.cc
```diff
@@ -20,6 +20,7 @@ enum { FIXED = -2, FREE_BUT_NOT_NUMBERED = -1 };
 
 template <class T>
 NumberingOf<T>::NumberingOf()
+  : FieldBase()
```
cwsmith: This appears to be the only non-whitespace change.  Would you please remove the whitespace changes below?

apf/apfNumbering.h
```diff
@@ -48,27 +48,44 @@ void destroyNumbering(Numbering* n);
   * \param node the node number withing the mesh entity
   * \param component the component number within the nodal tensor
   */
-void fix(Numbering* n, MeshEntity* e, int node, int component, bool fixed);
+void fix(Numbering * n,
```
cwsmith: Only whitespace changes in this file?

apf/apfAggregateNumbering.cc
```diff
@@ -0,0 +1,365 @@
...
+    //  contributes to the above scan
+    lcl_offset -= (strd * dofs_per_blk * blks_per_nd);
+    frst += lcl_offset;
+    // strds and frsts needs to be resized/updated
+    //  since we can have adjacent parts that were
+    //  not in the aggregation scope under which
+    //  init was called
+    int sz = PCU_Comm_Peers();
+    frsts.setSize(sz);
+    MPI_Allgather(&frst,1,MPI_INTEGER,
```
cwsmith: I don't recall if PCU provides `Allgather`.  If not, we should consider adding it to avoid/minimize direct MPI calls.  (repeat for the other `Allgather` calls)

apf/apfAggregateNumbering.cc
```diff
@@ -0,0 +1,365 @@
...
+      return new_ownr;
+    }
+    virtual void getCopies(MeshEntity * e, CopyArray & copies)
+    {
+      shr->getCopies(e,copies);
+      for(int ii = 0; ii < copies.getSize(); ++ii)
+      {
+        int new_ownr = -1;
+        if(new_cm != MPI_COMM_NULL)
+          MPI_Group_translate_ranks(pcu_grp,1,&copies[ii].peer,
```
cwsmith: Can we add a wrapper call to PCU for this translation?

apf/apfAggregateNumbering.cc
```diff
@@ -0,0 +1,365 @@
...
+      , pcu_grp()
+      , new_cm(ncm)
+      , new_grp()
+    {
+      if(pcu_cm == MPI_COMM_NULL)
+        pcu_cm = PCU_Get_Comm();
+      new_cm = PCU_Get_Comm();
+      MPI_Comm_group(pcu_cm,&pcu_grp);
+      if(new_cm != MPI_COMM_NULL)
+        MPI_Comm_group(new_cm,&new_grp);
```
cwsmith: Could we easily provide PCU wrappers for the group function and  `MPI_Group_translate_ranks`?

tobinw:
Sure thing I'll resolve that momentarily.

tobinw:
Resolved the conflicts, still need to add a test case for the PCU version of Allgather I added before this can be merged.

apf/apfNumbering.cc
tobinw: resolved

cwsmith:
I hit a few `unused-parameter` and `unused-but-set-variable` compilation warnings with gcc7.3 on rhel7.  A patch to silence them is attached.
[silenceWarnings.patch.gz](https://github.com/SCOREC/core/files/2375198/silenceWarnings.patch.gz)

wrtobin:
Once PR #234  is merge I am going to start looking at getting this functionality in as well, at least supporting/exposing implicit fields/numberings so we can supply closed-form representations of the fields/numberings under the hood and still use the 'read-only' version that results for various purposes.

Issue #159, opened 2018-07-24
apf::Field Reductions on Partition Boundaries


JaredCrean2:
I have an apf::Field where each partition writes values to the field (including the mesh entities on the partition boundary that it might not own), and I want to make the field consistent in parallel by taking the *minimum* of the values on all partitions.  `apf::accumulate` almost does this, except that it sums rather than takes the minimum.  It looks like [accumulateFieldData](https://github.com/SCOREC/core/blob/master/apf/apfFieldData.cc#L84) could be modified to take a reduction operation as an argument.

I need something like this for my current research, so I am willing to do the work on this.  If you are amenable to generalizing `apf::accumulate` in this way, I can spell out my proposed approach in greater detail.

JaredCrean2:
@cwsmith what do you think?

cwsmith:
From a brief review of the code it sounds like we should define an `apfSharedReduction` that takes the different operators and them implement the current `accumulate` API using it.  Is that what you had in mind?

JaredCrean2:
That's exactly what I was thinking.  Prototype implementation is [here](https://github.com/SCOREC/core/pull/160)

It defines an abstract `ReductionOp` type and then a few implementations for sum, min, and max.  I put these in `apf.h` because they need to be visible to functions in `apf.cc`, but they do clutter up the file.  Is there a better place for them?  They can't go in `apfFieldData.h` because then they wouldn't be visible in `apf.h` (`apf.h` is included by `apfFieldData.h`). 

The modified `accumulateFieldData` into `reduceFieldData` by adding a 4th argument that takes a `ReduceOp` and applies it to all the values at a node.  This requires doing a little more parallel communication, but avoids the need to do a second communication to broadcast the result back to the non-owner processes.

PR #161, opened 2018-07-29
introduce reduceFieldData and ReductionOp

JaredCrean2:
implements #159, this time based on the right branch.


cwsmith:
This looks good.  Would you please also add a test driver that exercises the various reductions and does some sort of sanity check of the results?

JaredCrean2:
Added a test.  It writes a field and then performs the reduction, and compares against the exact result.  I checked all 3 reduction (sum, min, max).  The first one checks that the original value is multiplied by `n`, and for the other ones I added the MPI rank of the process to the field value and checked that the reduction picked the right one.

cwsmith:
`fieldReduce.cc` is missing.

JaredCrean2:
New pull request #165 .  I ran CTest on one of the workstations, and all tests pass.

PR #163, opened 2018-07-31
implement FieldData::clone()

### LINKED ISSUE #162: copy constructors for apf::mesh and apf::Numbering
I need to copy a mesh and numbering.

I did not find a convenient way of copying an apf numbering. Does anything exist? Or should I implement something and submit a pull request?

For the mesh I can use `apf::createMdsMesh(model, mesh)` which works for my use case, however it seems like there should be a function `apf::copyMesh` which either does the same, or copies all of the underlying fields as well.

timetravellers:
this adds an implementation to clone field data which is needed for
apf::convert to copy fields. closes #162

### DIFF
--- a/apf/apfArrayData.cc
+++ b/apf/apfArrayData.cc
@@ -79,6 +79,14 @@ class ArrayDataOf : public FieldDataOf<T>
     T* getDataArray() {
       return this->dataArray;
     }
+    virtual FieldData* clone() {
+      //FieldData* newData = new TagDataOf<double>();
+      FieldData* newData = new ArrayDataOf<T>();
+      newData->init(this->field);
+      copyFieldData(static_cast<FieldDataOf<T>*>(newData),
+                    static_cast<FieldDataOf<T>*>(this->field->getData()));
+      return newData;
+    }
 
   private:
     /* data variables go here */
--- a/apf/apfFieldData.cc
+++ b/apf/apfFieldData.cc
@@ -10,11 +10,6 @@ FieldData::~FieldData()
 {
 }
 
-FieldData* FieldData::clone()
-{
-  abort();
-}
-
 void FieldData::rename(const char*)
 {
   abort();
--- a/apf/apf.cc
+++ b/apf/apf.cc
@@ -468,6 +468,13 @@ Field* createUserField(Mesh* m, const char* name, int valueType, FieldShape* s,
   return makeField(m, name, valueType, 0, s, new UserData(f));
 }
 
+void updateUserField(Field* field, Function* newFunc)
+{
+  UserData* ud = dynamic_cast<UserData*>(field->getData());
+  // ud will be null if the data is not user data
+  if (ud) ud->setFunction(newFunc);
+}
+
 void copyData(Field* to, Field* from)
 {
   copyFieldData(from->getData(), to->getData());
--- a/apf/apf.h
+++ b/apf/apf.h
@@ -721,13 +721,21 @@ struct Function
   virtual void eval(MeshEntity* e, double* result) = 0;
 };
 
-/** \brief Create a Field from a user's analytic function.
-  \details This field will use no memory, has values on all
-  nodes, and calls the user Function for all value queries.
-  Writing to this field does nothing.
-  */
+/* \brief Create a Field from a user's analytic function.
+ * \details This field will use no memory, has values on all
+ * nodes, and calls the user Function for all value queries.
+ * Writing to this field does nothing.
+ * \warning if you copy the mesh with apf::convert you may want to use
+ * apf::updateUserField to update function that this field refers to. This is
+ * extremely important if the analytic function you use references user fields.
+ */
 Field* createUserField(Mesh* m, const char* name, int valueType, FieldShape* s,
     Function* f);
+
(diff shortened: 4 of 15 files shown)


apf/apfFieldData.cc
```diff
@@ -9,10 +12,13 @@ namespace apf {
 FieldData::~FieldData()
 {
 }
-
 FieldData* FieldData::clone()
 {
-  abort();
+  FieldData* newData = new TagDataOf<double>();
```
tobinw: Do we want to commit to all copied fields being created as mesh tags? 

Certainly all fields using the primary apf interfaces in apf.h are mesh tag fields but I'm not sure we want to have references to a subclass in a superclass. Should ```clone()``` be virtual and the type created depend on the subclass type?

apf/apfArrayData.cc
```diff
@@ -1,135 +1,133 @@
...
-    {
-      /* this retrieves all the data associated with (e) */
-      int first_node_index = getNumber(this->num_var,e,0,0);
-      int num_nodes = this->field->countNodesOn(e);
-      int num_components = this->field->countComponents();
-      int start = first_node_index*num_components;
-      for (int i=0; i<num_nodes*num_components; i++) {
-         data[i] = this->dataArray[start+i];
-      }
+  template <class T>
```
tobinw: is this all whitespace changes? 

If so revert these edits.

apf/apfFieldData.cc
timetravellers: This should be templated, so we don't have to use only a double field.

apf/apfArrayData.h
```diff
@@ -4,19 +4,32 @@
...
-
 #include "apfFieldData.h"
-
 namespace apf {
+  template <class T>
+  void freezeFieldData(FieldBase* base);
+  template <class T>
+  void unfreezeFieldData(FieldBase* base);
+  template <class T>
+  class ArrayDataOf : public FieldDataOf<T> {
```
tobinw: Why does this class need to be exposed? Is it used somewhere else in the commit?

apf/apfFieldData.cc
timetravellers: This is really up to you and @cwsmith. it was not pure virtual originally, so I just implemented it as that implied.

apf/apfArrayData.h
timetravellers: My mistake. I was using this in a previous attempt to get this working.

apf/apfFieldData.cc
```diff
@@ -1,6 +1,9 @@
 #include <PCU.h>
 #include "apfFieldData.h"
 #include "apfShape.h"
+#include "apfArrayData.h"
+#include "apf.h"
+#include "apfTagData.h"
```
tobinw: Still need to remove these.

tobinw:
@timetravellers This looks like a better approach to me, just need to remove a couple #include directives.

@cwsmith Anything jumping out at you? ```FieldData.h``` is non-public so changing the function to pure-virtual shouldn't break any downstream code and the only interface changes seems to be the additional function for ```UserData``` in ```apf.h```

timetravellers:
@tobinw and @cwsmith updated and squashed. Let me know if there is anything else needed before merge.

timetravellers:
@tobinw I am finally done with this. We at least clone all mesh data that is used in tobinw/biotissue

timetravellers:
The one possibility where this breaks downstream code is if the user copies fields/tags etc. on their own. The downstream user no longer needs to copy fields when using `apf::makeMdsMesh()` or `apf::convert()`.

cwsmith:
I hit a compile error on pachisi using gcc 7.3.0

```
/lore/cwsmith/develop/core/apf/apfConvert.cc: In member function ‘void apf::Converter::convertTag(apf::Mesh*, apf::MeshTag*, apf::Mesh*, apf::MeshTag*)’:
/lore/cwsmith/develop/core/apf/apfConvert.cc:249:22: error: ‘cerr’ is not a member of ‘std’
                 std::cerr << "Tried to convert unknown tag type\n";
                      ^~~~
/lore/cwsmith/develop/core/apf/apfConvert.cc:249:22: note: suggested alternative: ‘errc’
                 std::cerr << "Tried to convert unknown tag type\n";
                      ^~~~
                      errc
/lore/cwsmith/develop/core/apf/apfConvert.cc: In member function ‘void apf::Converter::convertTags()’:
/lore/cwsmith/develop/core/apf/apfConvert.cc:331:20: error: ‘cerr’ is not a member of ‘std’
               std::cerr << "Tried to convert unknown tag type\n";
                    ^~~~
/lore/cwsmith/develop/core/apf/apfConvert.cc:331:20: note: suggested alternative: ‘errc’
               std::cerr << "Tried to convert unknown tag type\n";
                    ^~~~
                    errc
make[2]: *** [apf/CMakeFiles/apf.dir/apfConvert.cc.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [apf/CMakeFiles/apf.dir/all] Error 2
make: *** [all] Error 2
```

timetravellers:
@cwsmith see if that fixes the issue.

PR #169, opened 2018-08-15
Add code coverage to nightly test

### LINKED ISSUE #60: add code coverage to nightly system

timetravellers:
This functionality is pulled over from my implementation in
tobinw/biotissue. Valgrind is used to do memory checking, and gcov is
used for coverage.

closes #60 

Note: I don't have an easy way to verify that this works. @seegyoung (I think you are in charge of the nightly tests) should try running this with `ctest --script nightly.cmake`. If all goes well the memory checking and test coverage should show up on cdash site.

### DIFF
--- a/cdash/nightly.cmake
+++ b/cdash/nightly.cmake
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 2.8)
 
 SET(CTEST_DO_SUBMIT ON)
+set(CTEST_DO_MEMCHECK ON)
+set(CTEST_DO_COVERAGE ON) 
 SET(CTEST_TEST_TYPE Nightly)
 
 set(CTEST_NIGHTLY_START_TIME "17:00:00 EST")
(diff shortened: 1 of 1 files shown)


cwsmith:
@timetravellers we have valgrind support that works with mpirun:
https://github.com/SCOREC/core/blob/ecc64db4e488b64d39a7439d50f711e8cdc7f1c6/test/testing.cmake#L5-L8
https://github.com/SCOREC/core/blob/a6b4cbf589e5ee64a5c5f88ab6a0e28a20da74a4/CMakeLists.txt#L61-L64
IIRC, the builtin cmake valgrind integration does not support being called by mpirun.

The gcov addition is interesting.  I'll check that out.  Does this address Dan's point about aggregate coverage?

I'm currently the nightly build maintainer.

timetravellers:
Valgrind seems to work with mpi and cdash . You can see the result of a dynamic analysis on my project here:
https://my.cdash.org/index.php?project=Biotissue&subproject=develop

As far as I know this does aggregate data as Dan was hoping for. The following project shows an example with coverage turned on (and working): https://open.cdash.org/index.php?project=CMake

On tobinw/biotissue and tobinw/amsi I had to turn off the code coverage because our codebase is not currently completely open (we include some Simmetrix calls).

PR #183, opened 2018-09-21
control print statment output

cwsmith:
This PR attempts to satisfy [PUMI's compliance](https://github.com/xsdk-project/xsdk-issues/issues/67) with [XSDK policy M11](https://figshare.com/articles/xSDK_Community_Package_Policies/4495136):

> M11​. No package should have hardwired print or I/O statements that cannot be turned off through a
programmatic interface ; output should never be hard-wired to stdout or stderr. It is recommended
that packages provide a way for users to turn on output and allow them to direct where it goes. Also,
packages may print to stdout by default but only on one process (i.e., root rank “0”). But packages
may also be completely silent by default (and require that users turn on outputting in the appropriate
way).

In the `lion` directory (previously was used for compressing io) is an interface for conditional print statements:
https://github.com/cwsmith/core/blob/7d3e2e2fc195568ecabe3990017e68e12e9dd7b0/lion/lionPrint.h

Please let me know what you think of this approach/design.  I'm open to suggestions.

### DIFF
--- a/apf/apf.cc
+++ b/apf/apf.cc
@@ -20,6 +20,7 @@
 #include <cstdio>
 #include <cstdlib>
 #include <pcu_util.h>
+#include <lionPrint.h>
 
 namespace apf {
 
@@ -353,7 +354,7 @@ double computeCosAngle(Mesh* m, MeshEntity* pe, MeshEntity* e1, MeshEntity* e2,
 
     cosAngle = computeCosAngleInTri(m, pe, e1, e2, Q);
   } else {
-    printf("The requested angle computation is not implemented. Aborting! \n");
+    lion_oprint(1,"The requested angle computation is not implemented. Aborting! \n");
     abort();
   }
   return cosAngle;
@@ -439,7 +440,7 @@ void sharedReduction(Field* f, Sharing* shr, bool delete_shr,
 
 void fail(const char* why)
 {
-  fprintf(stderr,"APF FAILED: %s\n",why);
+  lion_eprint(1,"APF FAILED: %s\n",why);
   abort();
 }
 
--- a/apf/apfConvert.cc
+++ b/apf/apfConvert.cc
@@ -7,6 +7,7 @@
 #include "apfNumbering.h"
 #include <map>
 #include <pcu_util.h>
+#include <lionPrint.h>
 #include <iostream>
 #include <cstdlib>
 
@@ -247,7 +248,7 @@ class Converter
                 outMesh->setLongTag(newFromOld[e], out, lngData);
                 break;
               default:
-                std::cerr << "Tried to convert unknown tag type\n";
+                lion_eprint(1,"Tried to convert unknown tag type\n");
                 abort();
                 break;
             }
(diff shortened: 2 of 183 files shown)


wrtobin:
This looks like a good c-style approach to me. Couple of thoughts while looking at the interface:

I like that we're returning the ```vprintf()``` return value or zero, which gives us a method to determine whether our statement was actually printed or not, though it does mean that we can't differentiate between the verbosity restricting the printing and something in ```vprintf()``` failing. 

So we may want a function like ```bool lion_will_print(int lvl)``` or ```int lion_get_verbosity()``` to determine whether the current verbosity level will allow us to output at all. This might be useful in a situation where a lot of output is going to happen in a loop and we don't want to have to make a whole bunch of calls to ```lion_oprint()``` if we can simply avoid the loop altogether?

Currently we can just directly look at ```lion_verbosity_level``` but I'm not sure that should be publicly visible. Also ```lion_stdout``` and ```lion_stderr```. Is there a rationale for leaving those visible outside the lion compilation scope?

Oh also do we have any documentation or recommendations for developers on what the verbosity levels should be? At least in the SCOREC/core libraries? Or are we leaving this open to the developers to decide. 

Is there any way we can generate a compile warning/error for developers writing code that directly calls ```printf()`` to enforce using this interface? We could ```#define printf deprecated``` either on the compile line or in a header file in the appropriate compilation scopes or something like that? Alternatively we could simply add a script for use by Travis CI that greps the code (aside from ```SCOREC/core/lion/```) for usage of ```printf()``` and fails the integration test if it finds any occurrences?

How are we handling C++-style stream output? I don't know how prevalent that is in the codebase.

cwsmith:
`int lion_get_verbosity()` was added:

https://github.com/SCOREC/core/pull/183/commits/87699cdf2f888ab85f08a2f330591a344c2038a0

The global vars were removed from the header:

https://github.com/SCOREC/core/pull/183/commits/60444b4546b32a9c0c6740ec71df82ce547bff8e
 
> Oh also do we have any documentation or recommendations for developers on what the verbosity levels should be?

The only guidance is that a setting of zero produces no output and that the quantity of output increases as the setting increases from 0. 

> Is there any way we can generate a compile warning/error for developers writing code that directly calls `printf()` to enforce using this interface? We could `#define printf deprecated` either on the compile line or in a header file in the appropriate compilation scopes or something like that? Alternatively we could simply add a script for use by Travis CI that greps the code (aside from `SCOREC/core/lion/`) for usage of `printf()` and fails the integration test if it finds any occurrences?

`lionPrint.h` was pretty much added everywhere.  Placing the `#define printf deprecated` there seems like a logical thing to do.  Combining this with a Travis CI grep sounds good to me.  The grep should also exclude the `test` directory.

> How are we handling C++-style stream output? I don't know how prevalent that is in the codebase.

It is relevant; the `pumi` dir and a few other random spots have `cout` calls.  Any ideas what the interface should look like?  The tricky bit seems to me is how to specify the verbosity level.  Should the user have to call an ostream constructor in which they set the verbosity level?  Something like ` lionOut out(1); out << "foo\n";`. Or should it be something like `std::precision`; e.g., ` lionOut << lion::verbosity(1) << "foo\n";` (Note, I have no idea how this one works in practice).

cwsmith:
I gave up on finding a convenient method to intercept/redirect `cout` and `cerr` and just replaced the handful of uses with calls to `lion_[o|e]print` with `stringstream` usage as necessary.

cwsmith:
I just squashed all the commits in preparation for the merge.

wrtobin:
This looks good to merge after the CI testing completes.

PR #191, opened 2018-10-18
integrate over non-mesh dimension

### LINKED ISSUE #190: integrate process only works on mesh dim which may be different from the analysis dim
The example here is a 1d beam/truss element in 3d space. We want to integrate over 1D elements. Currently I work around this issue by replicating the `Integrator::process(Mesh* m)` code locally and process each element.

To reduce code duplication I want to propose the following change in core, which leaves the current api in tact, but adds a default argument to the process call.

e.g.
 
`Integrator::process(Mesh* m, int dim=-1)`

if the dim is `-1` the current behavior is maintained, otherwise we integrate over the user supplied dimension.

@cwsmith @wrtobin if you have no issues with this approach I will submit a pull request.

jacobmerson:
closes #190


wrtobin:
This looks good to me. 

Offhand @cwsmith do you know if `apf::Mesh::Integrate()` is actually covered by any of the automated test cases?

cwsmith:
Good question.  I don't know. @jacobmerson ?

wrtobin:
Yeah if not this is a decent opportunity to add something basic to cover the existing functionality and the new functionality.

jacobmerson:
I don't think there are any test cases for the integrator currently. Do you have any immediate thoughts on what a decent test case would be?

wrtobin:
We could just load a mesh with a known number of entities and create a counting integrator to count mesh elements and verify the correct number were visited?

wrtobin:
If the integrator only processes owned mesh entities we should be sure to include that consideration in the test case for a simple parallel mesh.

jacobmerson:
Ok. I will think about this tonight and try to put something together.

mds/apfMDS.h
```diff
@@ -53,7 +53,9 @@ typedef std::map<int, MeshEntity*> GlobalToVert;
 Mesh2* makeEmptyMdsMesh(gmi_model* model, int dim, bool isMatched);
 
 /** \brief load an MDS mesh and model from file
-  \param modelfile will be passed to gmi_load to get the model */
+  \param modelfile will be passed to gmi_load to get the model
+  \note gmi_register_mesh and gmi_register_null need to be called before this function
```
cwsmith: `gmi_register_sim` may also be called to enable loading of GeomSim, Parasolid, and ACIS models

jacobmerson:
@cwsmith I think I address your issue. Is there anything else that needs to be changed before merge?

cwsmith:
@jacobmerson looks good to me.

wrtobin:
@cwsmith I'm gonna merge this in at 4pm today unless anyone has any objections?

wrtobin:
I'm going to merge this now so I don't forget. It is a small enough change that I don't predict any immediate issues resulting, and reverting it in the case an issue does arise should be straightforward.

PR #200, opened 2018-12-13
Extensible vtk

JaredCrean2:
This PR make the `isPrintable` function available to users of apf.  I needed this to implement my own version of `populateWriteFields` using a different rule for which fields to print or not.

This is the first time I have created a new header file for Pumi, so I copied the copyright blurb from another file.  Someone should double check that I did it correctly.


JaredCrean2:
@cwsmith When you get a chance, could, you review?

cwsmith:
Please target this PR towards the develop branch.

JaredCrean2:
I don't need to extend the class.  The purpose of putting it in the header file was to share it between `apf` and `crv` (the class definition was duplicated previously).  The new header does not get installed, so I don't think users can extend it even if they wanted to.

PR #202, opened 2018-12-18
compute the derivative of a matrix field

### LINKED ISSUE #201: Matrix Field Gradient and Divergence
I am working on implementing the gradient of a 3x3 matrix field as needed for one of my projects. I am open to recommendations for the layout of the resulting 3rd order tensor if anyone has recommendations or preferences.

jacobmerson:
closes #201


jacobmerson:
Yes. I am just cleaning it up.

jacobmerson:
@cwsmith My original plan for testing doesn't work because I am using this with an SPR field (on linear elements) which due to the fitting procedure makes things not exactly match the analytic function...So I need to rethink a bit.

Is there testing for the other gradients? I'd like to use the same format as those tests.

jacobmerson:
@cwsmith I added a test for this...

PR #204, opened 2019-01-03
instatiate edge ip shape functions for getIPShape()

jacobmerson:
This just instatieates the edge shape functions so an IPShape function
can be created on an edge element.


cwsmith:
This is no longer needed.

PR #208, opened 2019-01-09
Jc copy data

JaredCrean2:
This change adds an optional argument to `apf::convert` (and functions that call `apf::convert`) that controls whether Fields/Numberings/GlobalNumberings/Tags are copied.  The default value is true, to preserve the original behavior.

This behavior is useful for doing an in-memory copy of the mesh data structure only.

### DIFF
--- a/apf/apfConvert.cc
+++ b/apf/apfConvert.cc
@@ -21,7 +21,7 @@ class Converter
       inMesh = a;
       outMesh = b;
     }
-    void run()
+    void run(bool copy_data=true)
     {
       createVertices();
       createEntities();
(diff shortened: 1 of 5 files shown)


JaredCrean2:
Fixed base branch and merged with recent changes

JaredCrean2:
@cwsmith When you get a chance, can you review this PR?

JaredCrean2:
I'm not sure how, because the most recent commit is a merge commit, so it doesn't show up with `git rebase -i'.  Is there another way of doing it?

cwsmith:
OK.  So it looks like you had to manually merge 6cc556e7ae4131e2ede0d53c48605a36ee9ed91c (or later) with your changes. Git rerere may have been able to resolve this without the merge commit, but I'm not sure:
https://git-scm.com/docs/git-rerere

PR #214, opened 2019-01-21
Sliver fix ignoring collapse

AjinkyaDahale:


cwsmith:
Is there any reason not to combine ('squash' in git terminology) these commits into a single commit?

Would you please add a test case that specifically exercises this code?

AjinkyaDahale:
Some commits can be squashed, but I believe it will make more sense to squash them into 2 or 3 commits than just one (removing the redundant call, new sliver-fix operations, and ignoring collapse).

Existing test cases for shape fix should cover the changes made here.

However, closing this PR for some time because there should be some changes I need to make before I can include this in develop.

PR #221, opened 2019-03-22
Some bugfixes and modifications to meshadapt

AjinkyaDahale:
Bugfixes:
* Prevent flags set for coarsening and refinement from hindering shape fixes
* Modified some compounds (face-split-collapse and single-edge-split-collapse) so they do not take into consideration pre-split elements while collapsing. Doing so would make the collapse create a degenerate element and fail.

Modification:
* Added a separate compound to handle flat elements that have one vertex close to an edge


AjinkyaDahale:
Accidentally PR'ed to master instead of develop

mortezah:
@AjinkyaDahale 
It seems that only 73 of the tests have been tried here https://travis-ci.org/SCOREC/core/builds/510077584?utm_source=github_status&utm_medium=notification

Please make sure that all the tests pass and then you can merge this to the develope branch.

AjinkyaDahale:
Confirmed that 122/122 tests pass. Going ahead with the merge.

PR #224, opened 2019-04-04
obtain connectivity array with edges embedded in higher dimensions

jacobmerson:
The test added in this commit depends on https://github.com/SCOREC/pumi-meshes/pull/1


jacobmerson:
Pulled to wrong branch.

PR #234, opened 2019-07-01
Initial implementation of support for complex fields

wrtobin:
Need to:
 - add a test case to CTest
 - squash the commits once the test is passing

Just wanted to get eyes on this to see if anyone can spot something I'm breaking or should do differently that isn't obvious to me.

### DIFF
--- a/apf/CMakeLists.txt
+++ b/apf/CMakeLists.txt
@@ -29,6 +29,7 @@ set(SOURCES
   apfVectorElement.cc
   apfVectorField.cc
   apfPackedField.cc
+  apfComplexField.cc
   apfNumbering.cc
   apfMixedNumbering.cc
   apfAdjReorder.cc
@@ -73,6 +74,7 @@ set(HEADERS
   apfField.h
   apfFieldData.h
   apfNumberingClass.h
+  apfComplex.h
 )
 
 # Add the apf library
--- a/apf/apf.cc
+++ b/apf/apf.cc
@@ -476,7 +476,7 @@ void unfreeze(Field* f)
 
 bool isFrozen(Field* f)
 {
-  return f->getData()->isFrozen();
+  return isFrozen(static_cast<FieldBase*>(f));
 }
 
 Function::~Function()
--- a/apf/apf.cc
+++ b/apf/apf.cc
@@ -13,6 +13,7 @@
 #include "apfMatrixField.h"
 #include "apfMatrixElement.h"
 #include "apfPackedField.h"
+#include "apfComplexField.h"
 #include "apfIntegrate.h"
 #include "apfArrayData.h"
 #include "apfTagData.h"
@@ -73,6 +74,8 @@ Field* makeField(
     f = new MatrixField();
   else if (valueType == PACKED)
     f = new PackedField(components);
+  else if (valueType == COMPLEX_PACKED)
+    f = new ComplexPackedField(components);
   else
     fail("invalid valueType in field construction\n");
   f->init(name,m,shape,data);
--- a/apf/apf.h
+++ b/apf/apf.h
@@ -133,6 +133,8 @@ enum ValueType {
   MATRIX,
  /** \brief a user-defined set of components */
   PACKED,
+ /** \brief a user-defined set of complex-components */
+  COMPLEX_PACKED,
  /** \brief placeholder used to set array sizes */
   VALUE_TYPES
 };
@@ -704,6 +706,9 @@ bool isPrintable(Field* f);
   */
 void fail(const char* why) __attribute__((noreturn));
 
+
+
+
 /** \brief Convert a Field from Tag to array storage. */
 void freeze(Field* f);
 
@@ -713,12 +718,15 @@ void unfreeze(Field* f);
 /** \brief Returns true iff the Field uses array storage. */
 bool isFrozen(Field* f);
 
+template <typename T>
+T* getArrayDataT(Field * f);
+
 /** \brief Return the contiguous array storing this field.
   \details This function is only defined for fields
   which are using array storage, for which apf::isFrozen
   returns true.
  */
-double* getArrayData(Field* f);
+inline double * getArrayData(Field* f) { return getArrayDataT<double>(f); }
 
 /** \brief Initialize all nodal values with all-zero components */
 void zeroField(Field* f);
--- a/apf/apfComplex.h
+++ b/apf/apfComplex.h
@@ -1,0 +1,14 @@
+#ifndef APFCOMPLEX_H_
+#define APFCOMPLEX_H_
+
+#ifdef C_COMPLEX
+  #include <complex.h>
+#endif
+
+#define CXX_COMPLEX 1
+#ifdef CXX_COMPLEX
+  #include <complex>
+  using double_complex = std::complex<double>;
+#endif
+
+#endif
--- a/pumi/pumi_ghost.cc
+++ b/pumi/pumi_ghost.cc
@@ -400,7 +400,7 @@ void pumi_ghost_create(pMesh m, Ghosting* plan)
   std::vector<apf::Field*> frozen_fields;
   for (int i=0; i<m->countFields(); ++i)
   {
-    pField f = m->getField(i);
+    apf::Field * f = m->getField(i);
     if (isFrozen(f))
     {
       frozen_fields.push_back(f); // turn field data from tag to array
--- a/apf/apfArrayData.cc
+++ b/apf/apfArrayData.cc
@@ -1,4 +1,5 @@
 #include "apfArrayData.h"
+#include "apfComplex.h"
 #include "apfNumbering.h"
 #include "apfTagData.h"
 
@@ -125,19 +126,29 @@ void unfreezeFieldData(FieldBase* field) {
 }
 
 /* instantiate here */
+template void freezeFieldData<double_complex>(FieldBase* field);
 template void freezeFieldData<int>(FieldBase* field);
 template void freezeFieldData<double>(FieldBase* field);
+template void unfreezeFieldData<double_complex>(FieldBase * field);
 template void unfreezeFieldData<int>(FieldBase* field);
 template void unfreezeFieldData<double>(FieldBase* field);
 
-double* getArrayData(Field* f) {
-  if (!isFrozen(f)) {
+template <typename T>
+T* getArrayDataT(FieldBase* f)
+{
+  if(!isFrozen(f))
     return 0;
-  } else {
-    FieldDataOf<double>* p = f->getData();
-    ArrayDataOf<double>* a = static_cast<ArrayDataOf<double>* > (p);
+  else
+  {
+    FieldDataOf<T>* p = reinterpret_cast<FieldDataOf<T>*>(f->getData());
+    ArrayDataOf<T>* a = static_cast<ArrayDataOf<T>*>(p);
     return a->getDataArray();
   }
 }
+
+template double_complex* getArrayDataT(FieldBase* field);
+template int* getArrayDataT(FieldBase* field);
+template double* getArrayDataT(FieldBase* field);
+
 
 }
--- a/apf/apfComplexField.cc
+++ b/apf/apfComplexField.cc
@@ -1,0 +1,21 @@
+#include "apfComplexField.h"
+#include "apfElement.h"
+
+namespace apf {
+
+Element * ComplexPackedField::getElement(VectorElement * e)
+{
+  return new Element(this,e);
+}
+
+void ComplexPackedField::project(Field*)
+{
+  fail("ComplexPackedField::project unimplemented");
+}
+
+void ComplexPackedField::axpy(double, Field*)
+{
+  fail("ComplexPackedField::axpy unimplemented");
+}
+
+}
--- a/apf/apfComplexField.h
+++ b/apf/apfComplexField.h
@@ -1,0 +1,25 @@
+#ifndef APFCOMPLEXFIELD_H_
+#define APFCOMPLEXFIELD_H_
+
+#include "apfField.h"
+#include "apf.h"
+
+namespace apf
+{
+class ComplexPackedField : public Field
+{
+public:
+  ComplexPackedField(int c) : components(c) {}
+  virtual ~ComplexPackedField() {}
+  virtual Element * getElement(VectorElement*);
+  virtual int getValueType() const { return COMPLEX_PACKED; }
+  virtual int countComponents() const { return components; }
+  virtual void project(Field  * frm);
+  virtual void axpy(double a, Field * x);
+private:
+  int components;
+};
+
+}
+
+#endif
--- a/apf/apfField.cc
+++ b/apf/apfField.cc
@@ -126,4 +126,9 @@ void zeroField(Field* f)
   op.apply(f);
 }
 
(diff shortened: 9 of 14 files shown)


apf/apfComplex.h
```diff
@@ -0,0 +1,14 @@
+#ifndef APFCOMPLEX_H_
+#define APFCOMPLEX_H_
+
+#ifdef C_COMPLEX
+  #include <complex.h>
+#endif
+
+#define CXX_COMPLEX 1
```
wrtobin: Oh yeah, need to add this as a CMake option rather that just hard defining it to be CXX_COMPLEX for now.

wrtobin:
Oh yeah I need to add a couple functions to the apf Simmetrix mesh class. 

@cwsmith aside from mds and simmetrix do we wrap any other mesh types that I'll need to modify either in CORE or downstream in associated projects?

I could swap the functions from being pure-virtual to just virtual with a default that fail("Unimplemented")-s

pumi/pumi_ghost.cc
```diff
@@ -400,7 +400,7 @@ void pumi_ghost_create(pMesh m, Ghosting* plan)
   std::vector<apf::Field*> frozen_fields;
   for (int i=0; i<m->countFields(); ++i)
   {
-    pField f = m->getField(i);
+    apf::Field * f = m->getField(i);
```
wrtobin: I should revert this, it crept in while I was working on an issue with freezing fields with FieldDataOf<T> where T != double.

apf/apf.h
```diff
@@ -704,6 +706,9 @@ bool isPrintable(Field* f);
   */
 void fail(const char* why) __attribute__((noreturn));
 
+
+
+
```
jacobmerson: why is there all this extra whitespace?

apf/apf.h
```diff
@@ -713,12 +718,15 @@ void unfreeze(Field* f);
 /** \brief Returns true iff the Field uses array storage. */
 bool isFrozen(Field* f);
 
+template <typename T>
+T* getArrayDataT(Field * f);
```
jacobmerson: Should this go into the cc file and use template specializations, as to not confuse the user space interface? If not possibly add a comment mentioning that this shouldn't be directly called by user?

pumi/pumi_ghost.cc
jacobmerson: I think you should keep the new version since it is consistent with the rest of the (non pumi) codebase

apf/apf.h
wrtobin: Yeah that's probably the way to go. Don't necessarily want users exposed to this template function. Wrapping up use of the template function in a couple of interface functions similar to how `getArrayData` is now implemented is probably the best short-term solution. 

Unfortunately there is a potential for misuse if a user supplies a 'Field' to one of these functions with the wrong underlying type, since we basically have to rely on a `reinterpret_cast` to get the underlying data of the different type. I'll see if I can add an assert or something to try to prevent this issue.

apf/CMakeLists.txt
```diff
@@ -73,6 +74,7 @@ set(HEADERS
   apfField.h
   apfFieldData.h
   apfNumberingClass.h
+  apfComplex.h
```
cwsmith: This and any other new public header will need to be added to the trililinos cmake source file list:
https://github.com/SCOREC/core/blob/master/apf/pkg_tribits.cmake

cwsmith:
> @cwsmith aside from mds and simmetrix do we wrap any other mesh types that I'll need to modify either in CORE or downstream in associated projects?

@mortezah Do we have a capstone implementation of apf_mesh?

wrtobin:
Switched base branch to develop cause I'm a dummy. Is there a way to restrict users from creating PRs for a branch (master in this case)?

mortezah:
Cameron,

Yes we have apf and gmi wrappers for capstone. But those are on a separate
private repo. I will take care of those changes if they are needed.

cwsmith:
>Is there a way to restrict users from creating PRs for a branch (master in this case)?

@wrtobin The only option I'm seeing in the github interface is to change the 'default' branch:

"The default branch is considered the “base” branch in your repository, against which all pull requests and code commits are automatically made, unless you specify a different branch."

I don't like this approach though as it will result in users who do a fresh clone getting placed into the `develop` branch instead of the stable `master` branch.

apf/apf.h
wrtobin: just crept into the commit, I've fixed this and once I squash the commits it will not be in the merge

wrtobin:
@cwsmith Well at least only a repo admin can screw it up (looking at me, @wrtobin)

Anyway I'm working on cleaning up the implementation and implementing a test case, but have a conceptual issue. 

While `FieldDataOf<T>` allows us to store fields with arbitrary data types (though is currently restricted to the Mesh::TagType enum types), the primary API is very much built around the `Field*` type which is assumed to contain a `FieldDataOf<double>`, though the actually-stored types might be any from the `apf::ValueType` enum which has SCALAR, VECTOR, MATRIX, PACKED (in the develop branch), which aren't types, they are intermediate data structures which all hold underlying `double` values. 

To implement the complex field I added a `COMPLEX_PACKED` value to that enum, which is passed through the typical field making APIs and results in the `Field*` having a `FieldDataOf<double_complex>` internally. 

The issue is somewhat twofold: (1) this new field violates the assumption that all `Field` objects store type double, (even though the underlying class hierarchy allows for this, the API is build around this assumption in some ways), and (2) adding `COMPLEX_PACKED` to the `ValueType` enum is changing the fundamental purpose of the enum from expressing the data structure used by the field to expressing the scalar type of the field. 

I'm trying to reconcile these issues and come up with the cleanest solution. Unfortunately since the API is built around the `Field->double` assumption, modifying it to allow other scalar types while maintaining the API might require some extensive changes 'under the hood' to do type checking (using a combination of `Field::getScalarType()`, templating, and `std::is_same<>()`, along with `reinterpret_cast`ing, all of which are not great implementation practices.

wrtobin:
Without a massive amount of work, the functionality we can get from a complex field would probably be:

create a complex field
freeze a complex field and retrieve the underlying array
access complex field data based on mesh adjacency information
unfreeze the field

Which is largely enough for the motivating use case in M3D-C1. 
It would also be nice to be able to build a complex `Element`

Assuming this all happens with the default `Field*` datatype as a 'wrapper' for the complex field, there is potential for misuse by users down the road if they build this field and try to use it in the general API. Many operations can be guarded with asserts or fails, but going though the entire API and checking that every time a `Field*` is passed around, the `getScalarType() == Mesh::DOUBLE` before using it is not desirable.

Do we have any mechanism to selectively publish/install headers only in specific 'developer' installations? Is having such a mechanism even a good idea or is it another giant can of worms? @cwsmith any thoughts?

apf/apf.cc
```diff
@@ -163,13 +161,15 @@ void destroyField(Field* f)
 
 void setScalar(Field* f, MeshEntity* e, int node, double value)
 {
+  PCU_DEBUG_ASSERT(f->getValueType() == SCALAR);
```
jacobmerson: If we are now allowing c++11 in CORE can this be done with a static assert?

apf/apf.h
```diff
@@ -718,15 +714,32 @@ void unfreeze(Field* f);
...
-template <typename T>
-T* getArrayDataT(Field * f);
+/** \brief Return the contiguous array storing this field.
+  \details This function is only defined for fields
+  which are using array storage, for which apf::isFrozen
+  returns true.
+  \note If the underlying field data type is NOT double,
+  this will cause an assert fail in all compile modes.
+ */
+double* getArrayData(Field* f);
```
jacobmerson: Do we have a mechanism for depreciation? If so, this should probably be depreciated and replaced with `getDoubleArrayData` to have a consistent API.

wrtobin: Since C++14 we have [[deprecated]], though this codebase is explicitly only up to C++11 IIRC and we haven't put in a deprecation mechanism as far as I know.

C++14 was basically just a small step from 11 so moving core to that should be possible, but is a different issue. 

For the time being I will introduce a more consistent API function and not in the doxygen this is deprecated (see apfNumbering.h near the bottom for the only other place this has been done I think).

apf/apf.cc
wrtobin: Core has been C++11 for a while IIRC. I've been using `auto` and range-based for's for a while no one has complained so far.

This assert has to be runtime though.

wrtobin:
Unfortunately the most effective way to get the desired functionality in without too much work cost is to basically replicate core portions of the field API for specifically complex fields since the `Field` type is explicitly supposed to contain only real-valued fields. 

Under the hood where all the templating is, things were easier to implement by, moving some things around, creating a few superclasses, and generalizing a few pieces of functionality, before unifying how the API uses the new structures to ensure the only real code duplication going on is the surface level API, since it is 100% the same except for some data types. In fact I kept the function names the same so porting existing algorithms can be a bit easier, just have to change the Field to ComplexField and any double arrays to double_complex.

Still need to write a robust test case and fix some bugs, I'm sure, along with adding the double_complex/std::complex<double> cmake option.

apf/apfComplex.h
```diff
@@ -11,4 +11,57 @@
...
+void destroyElement(ComplexElement* e);
+
+MeshElement* getMeshElement(ComplexElement* e);
+MeshEntity* getMeshEntity(ComplexElement* e);
+
+void getComponents(ComplexElement* e, Vector3 const& param, double_complex* components);
+int countNodes(ComplexElement* e);
+void getShapeValues(ComplexElement* e);
+void getShapeGrads(ComplexElement* e);
+void getPackedNodes(ComplexElement* e, NewArray<double_complex>& values);
```
wrtobin: I don't think these functions are implemented...

wrtobin: Now they are

apf/CMakeLists.txt
wrtobin: Just resolved this in 99423ccce83a9e925dba82bc72992e96a549d4f5

wrtobin:
Okay, everything should be done with this PR. All existing test cases are passing, including the new complex test case. We have a cmake option to determine which type of complex to use, and all tests pass in both cases.

wrtobin:
I'm looking into the travis CI issue and hope to resolve it soon.

wrtobin:
Okay now that the automated tests are passing I'm going to squash the commits to clean this up before we consider merging things in.

wrtobin:
Ah okay the squash will happen during the merge of this PR since we have force-pushing disabled on this repo (a smart move).

@cwsmith When you get a chance can you approve the PR/clear your review so I can merge?

CMakeLists.txt
```diff
@@ -84,6 +84,11 @@ endif()
 
 set(Trilinos_PREFIX "" CACHE STRING "Trilinos installation directory")
 
+option(CXX_COMPLEX "Use std::complex<double> instead of double_complex" OFF)
+if(CXX_COMPLEX)
+  add_definitions(-DCXX_COMPLEX)
```
cwsmith: It looks like `apfComplexType.h` is public so the `CXX_COMPLEX` compile flag will have to be passed along.  Alternatively, `apfComplexType.h` could be defined at compile time to avoid passing the compiler flag.

If we stick with the flag, should we prefix `CXX_COMPLEX` with `APF_` or `PUMI_` to avoid collisions?

apf/apfElementOf.h
```diff
@@ -14,6 +14,8 @@
 
 namespace apf {
 
+// std::enable_if_t<std::is_standard_layout<T>::value, bool> = 0 >
```
cwsmith: I assume this can be removed?

crv/crvQuality.cc
```diff
@@ -142,39 +142,39 @@ static double getTetPartialJacobianDet(apf::NewArray<apf::Vector3>& nodes,
  *
  */
 static double Nijk(apf::NewArray<apf::Vector3>& nodes,
-    int d, int I, int J)
+    int d, int II, int J)
```
cwsmith: What was the problem here with `I`?

apf/apfElementOf.h
wrtobin: Oh yeah my bad, that is related to type-safety for the field `valueType` access. I didn't quite get it working. I'll remove this.

crv/crvQuality.cc
wrtobin: When we compile with C-complex instead of C++-complex the "complex.h" header declares an `I` macro for the imaginary part of complex numbers in the global namespace... which is a hasstle. I could also `#undef I` in the `apfComplexType.h` header if that would be preferred, but that might have other issues?

CMakeLists.txt
wrtobin: Yeah I could just use cmake to process an apfComplexType.h.in file to create the apfComplexType.h file to avoid passing that along. I'll add a prefix as well, I think PUMI is preferable since we might wind up extending support outside of APF at some point.

wrtobin: Renamed PUMI_CXX_COMPLEX, setting this value to on results in `#define PUMI_CXX_COMPLEX` being placed in `apfComplexType.h`, otherwise it is `#define PUMI_C_COMPLEX` and we don't have to carry the compile flag forward.  

See a9ebda59d55e08f54ae856b2a4fe5962f5816441

apf/apfElementOf.h
wrtobin: removed in a9ebda59d55e08f54ae856b2a4fe5962f5816441

wrtobin:
Oh I still need to modify apfSIMMesh, let me see about that.

wrtobin:
So the additions to the simmetrix mesh type are fairly straightforward, except for the migration of complex tags, since simmetrix doesn't provide a general callback function for send/recv of tags in migration, only for integer tags and double tags so far as I can tell. Creating a new callback function would be feasible... if I knew enough about what the functions actually have to accomplish...

crv/crvQuality.cc
cwsmith: Ahhh.  Interesting.  The `#undef` seems messier than just avoiding `I`.

wrtobin:
Do we have anyone who knows about the simmetrix data migration callback functions? Or would it be best to submit a ticket?

Theoretically a user with a simmetrix mesh can still create complex fields, they just can't move meshes entities around at all...

cwsmith:
@wrtobin One of Onkar's students may have used it.  This example may be helpful:
https://www.scorec.rpi.edu/~cwsmith/SCOREC/SimModSuite_latest/PartitionedMesh/group__PMEx__Attach.html

wrtobin:
Yeah the issue is these functions:

```
  MD_setMeshCallback(id, CBmigrateOut, pm_sendDblArray, adc);
  MD_setMeshCallback(id, CBmigrateIn, pm_recvDblArray, adc);
```

We can get the signature of the pm_send/recvDblArray callback functions since that is in the documentation (and the same as the delete callback), but I don't precisely know what those callbacks are supposed to accomplish. Cursory searches through the relevant documentation doesn't come up with anything either. Unless we have someone who has written their own version of these callback functions I'm likely going to submit a low-level ticket with simmetrix before the end of the day.

cwsmith:
@wrtobin OK.  Let's submit a ticket.

PR #236, opened 2019-07-03
More general entity-type counting, expose the NoSharing type through the API for external usage

wrtobin:
move the implementation of NoSharing to be with the other sharing types

change it to be a singleton (which I usually hate but is actually warranted because NoSharing is basically just a stateless functor and we only ever need to retrieve a pointer to it externally) 

modify countEntitiesOfType to allow counting based on Sharings without changing the API semantics

add two tests to the test suite to test both a serial and partitioned mesh

### DIFF
--- a/apf/apf.cc
+++ b/apf/apf.cc
@@ -21,6 +21,7 @@
 #include "apfNumberingClass.h"
 #include <cstdio>
 #include <cstdlib>
+#include <PCU.h>
 #include <pcu_util.h>
 #include <lionPrint.h>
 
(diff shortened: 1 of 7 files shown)


wrtobin:
@cwsmith This PR should be fine to merge pretty much immediately.

wrtobin:
I'm actually going to add two API utility functions :`countLocalNodes()` and `countGlobalNodes()`, using this new functionality, so I'll add that in and add a test as well in a moment.

wrtobin:
@cwsmith That should wrap up this PR in terms of code additions.

 You want me to squash the commits into a single commit?

cwsmith:
@wrtobin Looks good.  I fine without the squash; its up to you.

wrtobin:
Okay I'll just merge it in that case, I want to create another small PR today and it's easier if this is in develop.

wrtobin:
Well, like a dummy I apparently set up this PR on master instead of develop.

Whats the easiest procedure to roll this back and get it in develop?

wrtobin:
Okay I merged a revert PR to master, glad GitHub has that. I'll make a new PR for the branch to go to develop.

PR #238, opened 2019-07-03
Expose NoSharing type through a simple API for external usage. Implement some more general entity and node counting APIs.

wrtobin:


wrtobin:
I'm not gonna merge this in until we're sure it won't cause any issues with the accidental merge and the revert of PR #236 (#237).

wrtobin:
Forgetting the test case .cc file would've broken everything anyway. At least Travis catches this when I PR on the correct branch....

PR #243, opened 2019-08-19
Maximum height function for triangles and tetrahedra

zhang-alvin:
Added maximum height functions for triangles and tetrahedra based on the minimum height functions. There is a lot of duplicated code, but to maintain backwards compatibility I opted for separate function calls unless someone can suggest something more elegant.

This functionality is being used for Proteus and the mesh adaptation routines.


mortezah:
@cwsmith could you please merge this pull request. It seems that I don't have the right access for merging. This is a relatively old pull request, if there is any problem with the merge let me know.

cwsmith:
@zhang-alvin Would you please change the target of this pull request to the `develop` branch?

PR #244, opened 2019-08-21
make ENABLE_ZOLTAN available to preprocessor

JaredCrean2:
This makes the `ENABLE_ZOLTAN` flag passed to CMake available to the C pre-processor (via a CMake generated header).  The motivation for this is trying to make code work with or without Zoltan, ie:

```
#ifdef ENABLE_ZOLTAN
  apf::Splitter* splitter = apf::makeZoltanSplitter(...);
#else 
  apf::Splitter* splitter = Parma_MakeRibSplitter(...);
#fi
# do stuff with splitter
```

### DIFF
--- a/zoltan/CMakeLists.txt
+++ b/zoltan/CMakeLists.txt
@@ -38,8 +38,14 @@ else()
   )
 endif()
 
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/apfZoltanConfig.h.in"
+               "${CMAKE_CURRENT_BINARY_DIR}/apfZoltanConfig.h")
+
+# Set include directory for building apf_zoltan target
+include_directories("${CMAKE_CURRENT_BINARY_DIR}")
+
 # Package headers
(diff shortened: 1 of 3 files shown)


cwsmith:
We should probably call this `PUMI_HAS_ZOLTAN` or something like that to prevent conflicts with other packages that have a macro/pre-processor definition for zoltan.

JaredCrean2:
Seems reasonable, although I may not get to this for a few weeks (I'm supposed to defend 2 weeks from today).

cwsmith:
This was added as part of #344 .

bobpaw:
@cwsmith I think this can be closed.

cwsmith:
Yes,  thank you. 

The pre-processor  flag `PUMI_HAS_ZOLTAN` was added here:
https://github.com/SCOREC/core/pull/342/commits/3943ce4f6cc2fcd89adeb30d9903db18a756aad1

Issue #272, opened 2019-11-05
Global numberings are leaked on mesh destruction


jacobmerson:
When the mesh is destroyed it leaks the `GlobalNumberings`. The solution to this should be relatively straight forward (destroy the GlobalNumberings when the mesh is destroyed). However, this causes a segfault/invalid memory access in some cases.

The two tests that exercise the sefault well are `verify_convert`, and `mixedNumbering`.

PR #274, opened 2019-11-05
Partial solution for issue #272

jacobmerson:
This commit shows at a high level how the memory leak should be expected
to be fixed. It however causes a segfault for some test cases
(verify_convert, and mixedNumbering).

The segfault seems to come from mds trying to delete the same tags
multiple times.

### DIFF
--- a/apf/apfNumbering.cc
+++ b/apf/apfNumbering.cc
@@ -46,10 +46,12 @@ void NumberingOf<T>::init(
 }
 
 template <class T>
-void NumberingOf<T>::init(Field* f)
+void NumberingOf<T>::init(Field* f, bool global)
 {
   field = f;
   std::string nm = f->getName();
+  if(global)
+    nm += "_global";
   nm += "_num";
   init(nm.c_str(),
       f->getMesh(),
@@ -123,8 +125,11 @@ Numbering* createNumbering(
 
 void destroyNumbering(Numbering* n)
 {
-  n->getMesh()->removeNumbering(n);
-  delete n;
+  if(n)
+  {
+    n->getMesh()->removeNumbering(n);
+    delete n;
+  }
 }
 
 void fix(Numbering* n, MeshEntity* e, int node, int component, bool fixed)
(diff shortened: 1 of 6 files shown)


jacobmerson:
@cwsmith I'm a bit stuck on the solution here. Any chance you have an idea how to get around the double deletion of tags that come in if we delete both types of numberings?

jacobmerson:
I surprisingly figured this one out...the names of the tag data are super important for MDS, so if there are repeats nasty stuff can happen. Repeats were happening if you made `Numbering` and `GlobalNumbering` both initialized with a field.

apf/apfNumbering.cc
```diff
@@ -46,11 +46,19 @@ void NumberingOf<T>::init(
...
   std::string nm = f->getName();
+  if(global)
+    nm += "_global";
   nm += "_num";
+  // FIXME: Using this string here is a use after free error
+  // and will lead to undefined behavior since the string gets
+  // destructed at the end of the function, and so should the
+  // underlying c_string. We should allocate a string to the
+  // heap and properly take care of the destruction of the string
+  // in the field base
```
cwsmith: As I understand,  this is OK since the `string` assignment operator used here:
https://github.com/SCOREC/core/blob/f7b18849c9e2ee82daad44aa8696c27206556fba/apf/apfField.cc#L20
copies the contents of the `const char*`.
http://www.cplusplus.com/reference/string/string/operator=/

cwsmith:
Does this completely fix the memory leak described in #272?

apf/apfNumbering.cc
jacobmerson: Indeed. For some reason I thought we were copying into a char *, not a string hence the confusion. I will remove the comment.

jacobmerson:
This fix does completely fix issue #272.

The caveat is that if you use the `createNumbering` api that takes a name and you construct a global, and local numbering with the same name it will cause issues. This issue is true for all of the tag based field types as mds uses names to find if tags already exist.

cwsmith:
OK.  Thank you.

Issue #275 was created to discuss/fix the name conflict problems.

Issue #285, opened 2020-01-06
SCOREC compilation error due to changes to STK Trilinos package


ikalash:
SCOREC failed to compile in the Albany nightlies last night due to a change in the STK Trilinos package.  The old versions of the set_cell_topology and get_cell_topology routines have been deprecated.  There are alternates in Trilinos/packages/stk/stk_mesh/stk_mesh/base/MetaData.hpp; see discussion at the end of this issue https://github.com/trilinos/Trilinos/issues/6522 .  Could you please fix the problem.  @alanw0 is a STK developer and should be able to help if you have any issues/questions.

cwsmith:
@alanw0 Do these changes https://github.com/SCOREC/core/pull/286/files make sense?  I'm new to STK.

ikalash:
Thanks for doing this so fast, Cameron!  Your changes look good to me and I was able to compile the code with this branch of SCOREC.  We'd need all of the changes working in Albany in order to test that everything is good now.  Hopefully I'll have something ready tomorrow.

cwsmith:
Thank you @ikalash.  The fixes have been merged to `develop` and should be in `master` by tomorrow morning.

ikalash:
Great, thanks for fixing it @cwsmith !

ikalash:
Seems this is still not in master.  @cwsmith , any idea on when it will get merged into master?  It is will take awhile, I might switch the nightlies to use develop for the time being.

cwsmith:
@ikalash It is there now:
https://github.com/SCOREC/core/commits/master

ikalash:
Oh ok great - thanks!

PR #286, opened 2020-01-07
update stk api for 12/20/2019 stk snapshot

cwsmith:
see #285

### DIFF
--- a/stk/apfMeshSTK.cc
+++ b/stk/apfMeshSTK.cc
@@ -205,9 +205,8 @@ static void declarePart(StkModel* model,
     const shards::CellTopology topo,
     StkMetaData* meta)
 {
-  stk::mesh::Part& part = meta->declare_part(model->stkName, rank);
   stk::topology t = stk::mesh::get_topology(topo);
-  stk::mesh::set_topology(part, t);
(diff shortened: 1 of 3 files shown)


stk/apfMeshSTK.cc
```diff
@@ -202,11 +202,12 @@ static void buildNodes(
...
 static void declarePart(StkModel* model,
     stk::mesh::EntityRank rank,
-    const CellTopologyData* topo,
+    const shards::CellTopology topo,
     StkMetaData* meta)
 {
   stk::mesh::Part& part = meta->declare_part(model->stkName, rank);
-  stk::mesh::set_cell_topology(part, topo);
+  stk::topology t = stk::mesh::get_topology(topo);
+  stk::mesh::set_topology(part, t);
```
alanw0: there is a 'declare_part_with_topology' method that might be more convenient, i.e., you could do this:
```
stk::topology t = stk::mesh::get_topology(topo);
stk::mesh::Part& part = meta->declare_part_with_topology(model->stkName, t);
```
then you wouldn't need the call to set_topology.

cwsmith: Thank you.  I made the change.

Issue #281, opened 2019-12-06
2D ugrid reader request


tuckerbabcock:
I'd like to request support for reading a 2D ugrid mesh. The documentation of the 2D ugrid format is here: http://www.simcenter.msstate.edu/software/downloads/doc/ug_io/2d_grid_file_type_ugrid.html
 
I'll be happy to work on this.

tuckerbabcock:
@cwsmith  I think I’ve finished modifying the ugrid reader to read the 2D mesh format, but after reading, I try to verify the mesh with: `ugrid_mesh->verify();`, which fails because I haven’t classified mesh vertices onto any model entities. The 2D ugrid format classifies both triangles/quads onto model faces, and the boundary edges onto model edges, yet doesn’t provide classification information for the mesh vertices. What am I to do? I’ve tried to follow the existing approach for reading a 3D ugrid mesh, but it doesn’t seem like that explicitly classifies mesh vertices. I suppose the classification of every vertex could be derived from the classification of adjacent edges, though this seems tedious.
Will I still be able to use MeshAdapt without this vertex classification?

Thanks

cwsmith:
The 3D ugrid reader creates a very simple model using the `deriveMdsModel(...)` API
https://github.com/SCOREC/core/blob/79032805c8c1ce90d661130df11b811387c24b4d/test/ugrid.cc#L60
described here:
https://github.com/SCOREC/core/wiki/Geometric-Models#gmi-topological-model

I don't see any mesh classification data being set in `mdsUgrid.cc`:
https://github.com/SCOREC/core/blob/59d03165714ac2c2069f4de6a3d08bd3ba041d20/mds/mdsUgrid.cc

Would you please share your code? A link to your fork/branch would be ideal.

tuckerbabcock:
Hi Cameron,

I've been working here: https://github.com/tuckerbabcock/core/blob/ugrid2D-dev/mds/mdsUgrid.cc

It's fairly rough right now, but I've added a load ugrid function with the signature
`Mesh2* loadMdsFromUgrid(gmi_model* g, const char* filename, const int mesh_dim)`,
which calls the existing loadMdsFromUgrid function if 3D, and if 2D calls a new function I wrote `readUgrid2D`. 

In functions called by `readUgrid2D` I iterate over and classify every mesh entity. To classify the vertices I've used the classification of the adjacent edges, done in `classifyVtx`. After I manually classified every entity the mesh passes verification.

[...]

PR #289, opened 2020-01-25
Add support for reading 2D ugrid files

tuckerbabcock:
This PR addresses #281 and extends `apf::loadMdsFromUgrid` to support reading 2D ugrid meshes. It sets the parametric values of mesh vertices for 2D meshes by inverse evaluating the geometric model at the coordinates of the vertex, if the `gmi` model supports the inverse evaluation.

**NOTE**: I've used C++11 features in the code that sets vertex parametric coordinates (the developer guide doesn't specify what maximum standard should be used).

### DIFF
--- a/mds/apfMDS.h
+++ b/mds/apfMDS.h
@@ -196,6 +196,9 @@ Mesh2* loadMdsFromGmsh(gmi_model* g, const char* filename);
 
 Mesh2* loadMdsFromUgrid(gmi_model* g, const char* filename);
 
+Mesh2* loadMdsFromUgrid(gmi_model* g, const char* filename,
+                        const int mesh_dim);
+
 void printUgridPtnStats(gmi_model* g, const char* ugridfile, const char* ptnfile,
     const double elmWeights[]);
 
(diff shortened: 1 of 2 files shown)


cwsmith:
@tuckerbabcock Thank you.  

Do you have a 2d ugrid file we can add to our nightly test meshes repo?

This can be tested by passing a 2d ugrid mesh file into the [from_ugrid](https://github.com/SCOREC/core/blob/70374171728094f4db7081b1b420205fc65b7d38/test/ugrid.cc) driver/utility?

tuckerbabcock:
@cwsmith yes I can add one, how should I add it? Does the mesh need to have a model associated with it?

cwsmith:
Thank you.  

I sent you an invite to 'collaborate' on the pumi-meshes repo.  

We currently have a 3d mesh in the `fun3d` dir.
https://github.com/SCOREC/pumi-meshes/tree/6d1601a29152e2043b16b2522c1a346b877920a0/fun3d

It probably makes sense to rename that directory `ugrid` and add the 2d mesh there.  In `core` we'll need to modify `test/testing.cmake` accordingly.

If you want to make the directory name change and upload the 2d mesh I will make the changes to `testing.cmake`.

tuckerbabcock:
I renamed that directory and added the 2D mesh.

PR #303, opened 2020-04-28
Nedelec

Samiullah-Malik:
Implements Nedelec Shape functions. These vector shape functions are used for Electromagnetic simulations. We need these to be able to transfer the field after the FEM solve and to perform error estimation.

PR #304, opened 2020-04-29
Nedelec (Re-based)

mortezah:
This is a resubmission of pull request https://github.com/SCOREC/core/pull/303 

The branch needed to be rebased since it was initially branched of `master` instead of `develop`

Implements Nedelec Shape functions. These vector shape functions are used for Electromagnetic simulations. We need these to be able to transfer the field after the FEM solve and to perform error estimation.


mortezah:
I realized some of the auto checks failed. By looking at it, it seems that a few of the tests cannot pass. They pass on my workstations. Is there something missing?

cwsmith:
Those few failing tests in Travis CI can be ignored; the `pumi_meshes` submodule was out of date. I pushed a fix to develop this morning.

Issue #298, opened 2020-03-23
mdsUgrid.cc breaks build on XL compiler


jacobmerson:
This code uses c++11 features which are not available in the XL compiler suite.

 Is there a preferred method we are using to only enable certain features if the appropriate compilers/language features exist?

jacobmerson:
For the specific case of XL I might be able to use qlanglvl to fix the error (will test shortly), however we should have a uniform way of dealing with code which requires certain compiler versions.

cwsmith:
This was on AiMOS?

jacobmerson:
I was compiling on summit, however I suspect the problem will be encountered anytime you use a compiler which doesn't have the c++11 standard enabled.

cwsmith:
On AIMOS, I built a minimal install of core with the following commands:

```
module load xl_r/16.1.1 spectrum-mpi
export OMPI_CXX=xlc++_r
export OMPI_CC=xlc_r

cmake3 ../core -DCMAKE_CXX_COMPILER=mpicxx -DCMAKE_C_COMPILER=mpicc
```

As described, this fails with:

```
[ 24%] Building CXX object mds/CMakeFiles/mds.dir/mdsUgrid.cc.o
cd /gpfs/u/home/MPFS/MPFSsmth/barn/build-core-dcs-xlSpectrum/mds && /opt/ibm/spectrum_mpi/bin/mpicxx -+  -I/gpfs/u/home/MPFS/MPFSsmth/barn/build-core-dcs-xlSpectrum/mds -I/gpfs/u/home/MPFS/MPFSsmth/barn/core/pcu -I/gpfs/u/home/MPFS/MPFSsmth/barn/core/pcu/reel -I/gpfs/u/home/MPFS/MPFSsmth/barn/core/gmi -I/gpfs/u/home/MPFS/MPFSsmth/barn/core/lion -I/gpfs/u/home/MPFS/MPFSsmth/barn/core/apf -I/gpfs/u/home/MPFS/MPFSsmth/barn/core/can -I/gpfs/u/home/MPFS/MPFSsmth/barn/core/mth  -O2 -g    -o CMakeFiles/mds.dir/mdsUgrid.cc.o -c /gpfs/u/home/MPFS/MPFSsmth/barn/core/mds/mdsUgrid.cc
/gpfs/u/home/MPFS/MPFSsmth/barn/core/mds/mdsUgrid.cc:319:35: error: expected expression
                                  [upward_dim](const int i) {
                                  ^
[...]

PR #305, opened 2020-05-04
Update CMake CXX11 flags and disable openmpi warning

jacobmerson:
The changes here are fairly small, just using the correct "CMake" way to deal with language definitions. This should be more portable especially when using a newer version of CMake.


cwsmith:
Thanks @jacobmerson.

Do you see any downside to making C++11 required?

If not, we should consider/discuss the changes here:
https://github.com/SCOREC/core/tree/c++11Required
.  Issue #298 discusses some ctest failures with the XL compiler.

jacobmerson:
I don't know if I'm the one qualified to discuss the downsides to enable c++11 by default...I think this really depends on the hardware we are expected to support. I.E. Is there hardware that core needs to run on which can't support c++11? I think most modern systems have support by now (even XL supports the full c++11 standard)

cwsmith:
Given #345 I think this can be closed.  If openmpi C++ interface warnings are still an issue we should create a new pr to fix it.

PR #307, opened 2020-05-18
Add region classification to apf::deriveMdlFromManifold

zhang-alvin:
PR to add mesh region to model region classification to the 2D/3D functions for deriving a discrete geometry based on mesh entity classifications. 

This is needed for the Proteus project to work correctly with multi-region domains.

Based on how the model is derived, there isn't a clear way to define the bounding surfaces for the model regions or how the regions are separated. I do not know how this will affect mesh adaptation.

### DIFF
--- a/mds/apfMDS.cc
+++ b/mds/apfMDS.cc
@@ -1156,6 +1156,7 @@ void deriveMdlFromManifold(Mesh2* mesh, bool* isModelVert,
 
 void derive2DMdlFromManifold(Mesh2* mesh, bool* isModelVert,
 			     int nBEdges, int (*bEdges)[4],
+                 std::vector<int> regionIDs,
 			     GlobalToVert &globalToVert,
 			     std::map<int, apf::MeshEntity*> &globalToFace)
 {
@@ -1216,6 +1217,13 @@ void derive2DMdlFromManifold(Mesh2* mesh, bool* isModelVert,
         mesh->setIntTag(vertsAdjToEdge[k], classifnTag, newTagData);
       }
     }
+  }
+
+  for (std::vector<int>::iterator it = regionIDs.begin() ; it != regionIDs.end(); ++it)
+  {
+    tagData[0] = 2;
+    tagData[1] = regionIDs[*it];
+    mesh->setIntTag(globalToFace[*it],classifnTag,tagData);
   }
 
   // TODO: Use classifnTag to classify
--- a/mds/apfMDS.cc
+++ b/mds/apfMDS.cc
@@ -1230,9 +1230,45 @@ void derive2DMdlFromManifold(Mesh2* mesh, bool* isModelVert,
 
   for(unsigned int i=0; i<regionIDs.size();i++)
   {
-    tagData[0] = 2;
-    tagData[1] = regionIDs[i];
+    tagData[0] = newTagData[0] = 2;
+    tagData[1] = newTagData[1] = regionIDs[i];
+
     mesh->setIntTag(globalToFace[i],classifnTag,tagData);
+
+    //check edge adjacency and classify if not classified already
+    mesh->getDownward(globalToFace[i], 1, edgesAdjToFace);
+    for (int k = 0; k <2; ++k) {
+        if (mesh->hasTag(edgesAdjToFace[k], classifnTag)) 
+        {
+            mesh->getIntTag(edgesAdjToFace[k], classifnTag, tagData);
+            if (tagData[0] > newTagData[0]) 
+            {
+	             mesh->setIntTag(edgesAdjToFace[k], classifnTag, newTagData);
+            }
+        } 
+        else 
+        {
+            // Edge has no classification, use the region's
+            mesh->setIntTag(edgesAdjToFace[k], classifnTag, newTagData);
+        }
+        
+
+        //get vertices in the meantime
+        mesh->getDownward(edgesAdjToFace[k], 0, vertsAdjToEdge);
+        for (int k = 0; k <2; ++k) {
+            if (mesh->hasTag(vertsAdjToEdge[k], classifnTag)) {
+	            mesh->getIntTag(vertsAdjToEdge[k], classifnTag, tagData);
+                if (tagData[0] > newTagData[0]) {
+	                mesh->setIntTag(vertsAdjToEdge[k], classifnTag, newTagData);
+                }
+            } else {
+                // Vertex has no classification, use the edge's
+                mesh->setIntTag(vertsAdjToEdge[k], classifnTag, newTagData);
+            }
+        }
+
+    }
+
   }
 
   // TODO: Use classifnTag to classify
--- a/mds/apfMDS.cc
+++ b/mds/apfMDS.cc
@@ -1237,7 +1237,7 @@ void derive2DMdlFromManifold(Mesh2* mesh, bool* isModelVert,
 
     //check edge adjacency and classify if not classified already
     mesh->getDownward(globalToFace[i], 1, edgesAdjToFace);
-    for (int k = 0; k <2; ++k) {
+    for (int k = 0; k < 3; ++k) {
         if (mesh->hasTag(edgesAdjToFace[k], classifnTag)) 
         {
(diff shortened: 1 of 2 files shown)


zhang-alvin:
initialClassification
postAdaptClassification

So it looks like some additional classification info will be needed to ensure that new elements have the same region classification as the parents.

mortezah:
> initialClassification
> postAdaptClassification
> 
> So it looks like some additional classification info will be needed to ensure that new elements have the same region classification as the parents.

is this a 2D example?

zhang-alvin:
> > initialClassification
> > postAdaptClassification
> > So it looks like some additional classification info will be needed to ensure that new elements have the same region classification as the parents.
> 
> is this a 2D example?

Yes. Do you have an idea of what might be needed to preserve the region classification in this context?

mortezah:
> > > initialClassification
> > > postAdaptClassification
> > > So it looks like some additional classification info will be needed to ensure that new elements have the same region classification as the parents.
> > 
> > 
> > is this a 2D example?
> 
> Yes. Do you have an idea of what might be needed to preserve the region classification in this context?

Can you have the meshes overlay on each other? I want to see what is the operator causing this.

zhang-alvin:
@mortezah 

overlay_withCoarsen

I've set the initial mesh edges to be colored white whereas the adapted mesh to be colored black. The dark blue patches seem to result from edge swaps and collapses.

mortezah:
> @mortezah
> 
> overlay_withCoarsen
> 
> I've set the initial mesh edges to be colored white whereas the adapted mesh to be colored black. The dark blue patches seem to result from edge swaps and collapses.

That does not seem to be always the case. When you do the adapt can you set in->shouldCoarsen = false, so we now that we only have refinement and swap happening?

mortezah:
One more question. Does this happen for 3D meshes?

mortezah:
One last question. If the new regions (the ones in dark blue in the pictures) are not classified on the correct regions, what are they classified on? Do they even have classifications?

zhang-alvin:
overlay_withNoCoarsen

Overlay with `in->shouldCoarsen=false`. I have not checked with 3D meshes but I suspect similar issues/problems will arise.

zhang-alvin:
> One last question. If the new regions (the ones in dark blue in the pictures) are not classified on the correct regions, what are they classified on? Do they even have classifications?

Good question. This output suggests that were classified on a region with ID 0, although no such region should exist in the original mesh.

zhang-alvin:
@mortezah , let me know if this will be simpler to explain/discuss via a slack call, if you have an idea of what might be happening.

mortezah:
> > One last question. If the new regions (the ones in dark blue in the pictures) are not classified on the correct regions, what are they classified on? Do they even have classifications?
> 
> Good question. This output suggests that were classified on a region with ID 0, although no such region should exist in the original mesh.

So when you define the classifications for a given mesh region to be on a given model region, what do you do for the bounding entities of that region? 

One more thought. Can you maybe try a two-region mesh that is generated with simmetrix (converted to pumi) and some made-up size field and see if the same thing happens after adapt?
I have a 2D mesh/model with 2 regions here https://www.dropbox.com/s/tegx5xh4gnmreey/mesh_model.zip?dl=1

zhang-alvin:
> > > One last question. If the new regions (the ones in dark blue in the pictures) are not classified on the correct regions, what are they classified on? Do they even have classifications?
> > 
> > 
> > Good question. This output suggests that were classified on a region with ID 0, although no such region should exist in the original mesh.
> 
> So when you define the classifications for a given mesh region to be on a given model region, what do you do for the bounding entities of that region?
> 
> One more thought. Can you maybe try a two-region mesh that is generated with simmetrix (converted to pumi) and some made-up size field and see if the same thing happens after adapt?
> I have a 2D mesh/model with 2 regions here https://www.dropbox.com/s/tegx5xh4gnmreey/mesh_model.zip?dl=1

I think the first thought regarding bounding entities of the model region is the likely cause of the issue. As of now, there is no such information being supplied. There is no information being propagated down to lower dimensional entities either, so a vertex in the model region does not have the right classification.

mortezah:
> > > > One last question. If the new regions (the ones in dark blue in the pictures) are not classified on the correct regions, what are they classified on? Do they even have classifications?
> > > 
> > > 
> > > Good question. This output suggests that were classified on a region with ID 0, although no such region should exist in the original mesh.
> > 
> > 
> > So when you define the classifications for a given mesh region to be on a given model region, what do you do for the bounding entities of that region?
> > One more thought. Can you maybe try a two-region mesh that is generated with simmetrix (converted to pumi) and some made-up size field and see if the same thing happens after adapt?
> > I have a 2D mesh/model with 2 regions here https://www.dropbox.com/s/tegx5xh4gnmreey/mesh_model.zip?dl=1
> 
> I think the first thought regarding bounding entities of the model region is the likely cause of the issue. As of now, there is no such information being supplied. There is no information being propagated down to lower dimensional entities either, so a vertex in the model region does not have the right classification.

That is also my suspicion. In general, any mesh entity has to have a __unique__ classification. More specifically, each mesh entity has to be classified on the lowest possible dimension model entity. In other words, if M<sub>i</sub><sup>d<sub>i</sub></sup> is classified on G<sub>j</sub><sup>d<sub>j</sub></sup> then d<sub>i</sub> <= d<sub>j</sub>

zhang-alvin:
Okay, I think I'll be able to classify the "interior" vertices with the appropriate region classification. This won't resolve the missing bounding entities for the regions.

Is it possible to bound a model region in the following manner: 
Classification

Currently, the derived model function creates model edges that need not be contiguous. So is it reasonable to bound region 0 with edges {10,5,1}.

mortezah:
> Okay, I think I'll be able to classify the "interior" vertices with the appropriate region classification. This won't resolve the missing bounding entities for the regions.
> 
> Is it possible to bound a model region in the following manner:
> Classification
> 
> Currently, the derived model function creates model edges that need not be contiguous. So is it reasonable to bound region 0 with edges {10,5,1}.

A proper model would never look like the one you have shown in the picture. That being said, you might be able to do that (i.e., bounding model region 0 with edges 10, 5, 1) and get the mesh adapt to work, but note that there is going to be a problem if you ever have a mesh edge that connects the top and bottom model edges (the ones shown as edge 5 in your picture).

Also, I don't think that is the issue for the swaps failing to classify things properly (at least for the swaps that are happening away from any boundary). For a mesh region that is completely inside a given model region (i.e., no intersection with any boundary of the model) one needs to classify that mesh region an all of its downward entities (in your 2D example verts and edges) on the same model region. That should fix the problem with interior swaps.  The ones on the boundary are a bit trickier.

zhang-alvin:
> > Okay, I think I'll be able to classify the "interior" vertices with the appropriate region classification. This won't resolve the missing bounding entities for the regions.
> > Is it possible to bound a model region in the following manner:
> > Classification
> > Currently, the derived model function creates model edges that need not be contiguous. So is it reasonable to bound region 0 with edges {10,5,1}.
> 
> A proper model would never look like the one you have shown in the picture. That being said, you might be able to do that (i.e., bounding model region 0 with edges 10, 5, 1) and get the mesh adapt to work, but note that there are going to be a problem. For example, if you ever have a mesh edge that connects the top and bottom model edges (the ones shown as edge 5 in your picture).
> 
> Also, I don't think that is the issue for the swaps failing to classify things properly (at least for the swaps that are happening away from any boundary). For a mesh region that is completely inside a given model region (i.e., no intersection with any boundary of the model) one needs to classify that mesh region an all of its downward entities (in your 2D example verts and edges) on the same model region. That should fix the problem with interior swaps. The ones on the boundary are a bit trickier.

Yes this was never really meant to be a fully functioning model. This started off primarily as a hack to get Proteus working with adaptivity since there isn't a proper way to generate a model. We're actually looking into the Capstone workflow, so that might be the long-term solution. I'll try the downward classifications to resolve the swap issue.

zhang-alvin:
Going to close this for now since it doesn't since adding downward classification in 3D isn't sufficient to avoid errors.

Issue #293, opened 2020-03-01
Curved mesh support along internal model interfaces?


tuckerbabcock:
Hi, I'm wondering if Bezier-curved meshes are supported for models where there internal faces? 

For example, consider this simple model of a box and cylinder intersecting, that share two faces: 
<img width="537" alt="Screen Shot 2020-02-29 at 9 23 13 PM" src="https://user-images.githubusercontent.com/19217934/75618304-33973600-5b3a-11ea-85b7-743cc0efd8ec.png">

I've generated a mesh on this, and am attempting to curve it, but get a `PCU_Assert_Fail` for `adjRegions->n <= 1` when querying the number of adjacent regions for one of the intersecting faces (which would have two adjacent regions). This leads me to believe that curved adaptation is not supported for a model like this?

The above model is an EGADS model, which I have been adding support for (on my fork), and it is entirely possible that there is an issue with my implementation that is causing this problem.

Thanks, Tucker

For reference, here is the relevant backtrace:
```
frame #0: 0x00007fff6b00c7fa libsystem_kernel.dylib`__pthread_kill + 10
frame #1: 0x00007fff6b0cebc1 libsystem_pthread.dylib`pthread_kill + 432
frame #2: 0x00007fff6af93a1c libsystem_c.dylib`abort + 120
frame #3: 0x0000000109dd8044 pumiAIM.so`PCU_Assert_Fail(msg="adjRegions->n <= 1 failed at /Users/tuckerbabcock/Developer/core/apf/apfMesh.cc + 254 \n") at pcu_util.c:16:3
frame #4: 0x0000000109c32268 pumiAIM.so`apf::Mesh::isParamPointInsideModel(this=0x0000000100790f00, g=0x0000000100790ea0, param=0x00007ffeefbf5128, x=0x00007ffeefbf5148) at apfMesh.cc:254:3
frame #5: 0x0000000109d2e9f1 pumiAIM.so`ma::interpolateParametricCoordinatesOnRegularFace(m=0x0000000100790f00, g=0x0000000100790ea0, t=0.5, a=0x00007ffeefbf5390, b=0x00007ffeefbf53a8, p=0x00007ffeefbf5568) at maSnap.cc:495:11
[...]

PR #330, opened 2021-02-12
Add support for EGADS geometry system

tuckerbabcock:
This PR adds support for using the EGADS geometry system (acdl.mit.edu/ESP). 

I've previously spoken to @cwsmith and @mortezah about the implementation, which is now fairly complete. Since this is my first fairly large PR to core, please let me know what else I should add to get this merged.


cwsmith:
Hi @tuckerbabcock .  Thank you for the contribution.  I think pumi users will like having an open source cad option.

Can you add some test input files for use with `test/split.c` to the https://github.com/SCOREC/pumi-meshes repo and add a call to split with those inputs to `test/testing.cmake` within a conditional that checks if EGADS is enabled.  Here is a similar example:
https://github.com/SCOREC/core/blob/d86c1ea4d3a7e2f27b733c578de52d6a2135b841/test/testing.cmake#L139-L146

Also, if possible, we should add a check on the version of EGADS that is found and compare that to a version range that we know works.  Note, this is assuming EGADS provides version info somehow and it can be accessed at compile time.  I can help with this as needed.

tuckerbabcock:
@cwsmith I'll work on adding a test that calls split with a mesh that uses an EGADS model. I'll also look into the version checks. 

Thanks for your feedback!

tuckerbabcock:
@cwsmith the EGADS header files `#define EGADSMAJOR` and `EGADSMINOR` that hold the version number. Should I check the version in the cmake files so it's done as part of the configure step? I think I might need your help to implement that

tuckerbabcock:
Hi @cwsmith,

I've attempted to add tests for `split`, and added mesh and model files to the pumi-meshes repo under a folder named `egads`. The tests start to run on my machine, but fail because it can't find the model file, even though if I've verified that it's looking in the correct place and the model file does exist there. Any thoughts?

apf/apfMesh.cc
```diff
@@ -242,18 +253,15 @@ bool Mesh::isParamPointInsideModel(ModelEntity* g,
...
+  (void)param;
+  (void)x;
+  return true; /*
   PCU_ALWAYS_ASSERT(adjRegions->n <= 1);
   gmi_ent* r = (gmi_ent*)adjRegions->e[0];
   gmi_eval(getModel(), (gmi_ent*)g, &param[0], &x[0]);
   int res = gmi_is_point_in_region(getModel(), r, &x[0]);
   gmi_free_set(adjRegions);
-  return (res == 1) ? true : false;
+  return (res == 1) ? true : false; */
```
jacobmerson, line 264: It looks like you break this function? Do we not test this?

gmi_egads/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/gmi_egads_config.h.in"
+               "${CMAKE_CURRENT_BINARY_DIR}/gmi_egads_config.h")
+
+#Sources & Headers
+set(SOURCES gmi_egads.c)
+set(HEADERS gmi_egads.h)
+
+add_library(gmi_egads ${SOURCES})
+
+target_include_directories(gmi_egads PUBLIC ${EGADS_INCLUDE_DIR})
```
jacobmerson, line 21: Cursory look over indicates that egads isn't part of the interface, so you should be using PRIVATE linkage and header file includes

gmi_egads/gmi_egads.c
```diff
@@ -0,0 +1,907 @@
...
+  int dim;
+  int tag;
+} egads_ent;
+
+/// global array of model entities
+/// egads_global_ents[0] - array of model verts
+/// egads_global_ents[1] - array of model edges
+/// egads_global_ents[2] - array of model faces
+/// egads_global_ents[3] - array of model regions
+static egads_ent *egads_global_ents[4];
```
jacobmerson, line 46: Why are you using global state? Won't this cause issues if you decide you want to read multiple egads files in the same program?

gmi_egads/gmi_egads.c
```diff
@@ -0,0 +1,907 @@
...
+/// adjacency table to be used for the "dummy" 3D elements that EGADS doesn't
+/// natively track
+static int **adjacency_table[6];
+
+/// read adjacency table from binary file
+static void read_adj_table(const char* filename,
+                           int *nregions)
+{
+  char *sup_filename;
+  sup_filename = EG_alloc(strlen(filename)+4+1);
```
jacobmerson, line 64: document why +4+1, also I don't think you free this

apf/apfMesh.cc
tuckerbabcock: These changes were based on the discussion in #293 and I haven't looked at them since. It looks like @mortezah may have implemented something that at least partially addresses the issue, so my changes could probably be reverted.

gmi_egads/gmi_egads.c
tuckerbabcock: Thanks for pointing this out... I have experienced the exact problem you mention where if I load multiple egads models I get a seg fault upon freeing the mesh for duplicate frees in this array. I hadn't given it much thought why, but I'll be sure to address this since I'm sure it's the reason.

gmi_egads/CMakeLists.txt
tuckerbabcock: I've got no objections to switching to private linkage, but I want to make sure that some of the symbols are not internal. While they're not part of the public interface for PUMI (and thus not in the header file), I use `struct gmi_model* gmi_egads_init(struct egObject *body, int numRegions)` and `void gmi_egads_init_adjacency(int ***adjacency)` in the code that generates a mesh in the PUMI format from an EGADS model, so I need those functions to be callable from an external project. I don't think I know enough about cmake to make that happen, or if that is even what you're really referring to.

tuckerbabcock:
Thanks for taking a look at the changes @jacobmerson. Your last point about the allocation for the `sup_filename` is related to a design decision I made that probably deserves a little more scrutiny.

For models that look like this, a box and cylinder intersecting that share two internal faces along the intersection:
image
The egads topology functions will report 1 region for this model, even though there are conceptually two distinct regions (the cylinder part and the box part). Further, since egads only sees one region, it can not tell you what model faces are adjacent to the cylinder, or any other adjacency relationships that involve a region.

To address this, during the mesh generation process I've written out a `model_name.egads.sup` "supplementary" model file that contains the region adjacency information. This file will be read during `gmi_egads_load` and will allow the `gmi_model` to be queried about the region adjacency relationships.

I have access to the region adjacency information during mesh generation through the CAPS interface which is part of the Engineering Sketch Pad that EGADS itself is a part of. CAPS uses analysis interface modules (AIMS) to generate meshes. I've written an AIM that will construct a PUMI mesh from the internal CAPS mesh description. The internal mesh description contains the expected association between mesh volume elements and the model region they belong to, which I'm able to query to build the adjacency table that is written to `model_name.egads.sup`.

The gist of this is: 
EGADS does not internally keep track of regions -> I have manually kept track of model regions and their adjacency relationships (stored in the supplementary model file) -> I figure out these adjacency relationships during the mesh generation process since the CAPS internal mesh representation contains an association between volume elements and region ID.

Any thoughts @cwsmith, @jacobmerson?

gmi_egads/gmi_egads.c
tuckerbabcock: Fixed in a7e41e3

gmi_egads/gmi_egads.c
tuckerbabcock: Fixed in 46f75a8

apf/apfMesh.cc
jacobmerson: I'm not familiar with the exact functionality, so @mortezah might be able to shed more light. But, based on the discussion the solution of `return true` looks like a work around which needs to get fully fixed before merging?

gmi_egads/CMakeLists.txt
jacobmerson: The `EGADS_INCLUDE_DIR` and `EGADS_LIBRARIES` point to the external egads libraries. If you are doing a public linkage/include you are requiring that the user of the compiled pumi library can also link against those libs. If you make the linkage private then you will need to have egads to compile pumi with egads support, but whoever consumes pumi will not need to have access to egads to compile/link. (although you will need the egads library at runtime if you do dynamic linkage)

Your functions `gmi_egads_init` and `gmi_egads_init_adjacency` are part of the `gmi_egads` library target that you create, and you want those visible to anyone who consumes your library, so adding the `CMAKE_CURRENT_SOURCE_DIR` and `include` as public targets is correct.

apf/apfMesh.cc
mortezah: @tuckerbabcock the quick fix of returning true you put in place at the time we were discussing issue #293 is obsolete now. The code has changed since then and the current implementation of this function should work for you case, as well. If it did not, let me know.

@jacobmerson

tuckerbabcock: Thanks @mortezah. I'll revert these changes.

gmi_egads/CMakeLists.txt
tuckerbabcock: Okay, I think I've addressed this in ad8ab1f.

jacobmerson: Looks good 😃

gmi_egads/gmi_egads.c
```diff
@@ -0,0 +1,832 @@
...
+  {
+    adj_faces = gmi_adjacent(m, adj_edges->e[i], 2);
+    for (int j = 0; j < adj_faces->n; ++j)
+    {
+      if (adj_faces->e[j] == to)
+      {
+        double t;
+        getVertexT(m, adj_edges->e[i], from, &t);
+        m->ops->reparam(m, adj_edges->e[i], &t, to, to_p);
+        goto cleanup;
```
cwsmith, line 266: Can the code at the `cleanup` marker be moved here?  Is the `goto` necessary?

gmi_egads/gmi_egads.c
```diff
@@ -0,0 +1,832 @@
...
+
+  free(egm);
+}
+
+static struct gmi_model_ops ops;
+
+/// TODO: Come up with a better flag? 
+/// TODO: re-write for EGADSlite - model loading is different
+// #ifdef HAVE_EGADS
+#if 1
```
cwsmith: IIUC, the generated cmake in this PR provides a preprocessor macro indicating if egads lite is available.

cwsmith (changes requested):
Looks good to me.  Thank you.  Sorry for the delay.

I left a few small comments in the code.

tuckerbabcock:
I've made some changes that enable to use of EGADSlite, the HPC friendly version of EGADS. The main limitation of EGADSlite is that it cannot load an EGADS model on its own. To work with this I've implemented that rank 0 will use EGADS to load the model, and then broadcast the model over MPI to all other ranks that will import it with EGADSlite (based on an example in the EGADS distribution that follows this process). To have gmi_egads use either EGADS or EGADSlite based on the MPI rank I've switched from linking EGADS into the gmi_egads library, and instead load either EGADS or EGADSlite at runtime with `dlopen`.

A potential headache with this new approach is that I've had to hardcode the function prototypes for all the EGADS APIs I use as function pointers that will be loaded with the correct symbols at runtime using `dlsym`. This means that if EGADS changes the signature of any of these functions it will be a runtime error instead of a compile-time error. I think I could address this in the future by re-writing gmi_egads in C++ and using `decltype` and `std::is_same`: https://stackoverflow.com/questions/14541519/is-it-possible-to-write-c-template-macros-to-check-whether-two-functions-have/14541544#14541544

For those curious, here's the paper describing EGADSlite: https://acdl.mit.edu/ESP/Publications/AIAApaper2018-1401.pdf

gmi_egads/gmi_egads.c
tuckerbabcock: This is outdated. I rewrote this for EGADSlite in the most recent commit.

cwsmith:
Re: https://github.com/SCOREC/core/pull/330#issuecomment-779377675

I think as long as the process (and code) used to generate the supplementary info is available then we should be fine.

Re: egads model loading + bcast

Thanks for the paper link.  From what I understand, this passage:

>  The initialization/model-read is oneof the differences, where anEGADSapplication prepares the data for use by the EGADSlite package.  This data can be written to disk or used live in a parallel setting. 

indicates that an EGADS API can be used to write a model that EGADSlite can load.  Could a pumi utility be created that reads the supplementary info and egads model and writes the egadslite model?  This seems like a simpler approach than broadcast + dlopen.

tuckerbabcock:
> indicates that an EGADS API can be used to write a model that EGADSlite can load.

Yes, that's correct. The function `EG_exportModel` is used to write the loaded EGADS model into a stream. An EGADSlite instance would then call `EG_importModel` to load the model. I think it would be pretty easy to make a pumi utility that could handle this conversion. And you're right, just using EGADSlite and eliminating the dependency on opencascade would make this code much more portable.

gmi_egads/gmi_egads.c
tuckerbabcock: I got rid of the `goto`. I'm not sure why I thought I needed it in the first place.

tuckerbabcock:
> I think as long as the process (and code) used to generate the supplementary info is available then we should be fine.

Yes, it's available here: https://github.com/tuckerbabcock/EngSketchPad/tree/beta

cwsmith:
> > indicates that an EGADS API can be used to write a model that EGADSlite can load.
> 
> Yes, that's correct. The function `EG_exportModel` is used to write the loaded EGADS model into a stream. An EGADSlite instance would then call `EG_importModel` to load the model. I think it would be pretty easy to make a pumi utility that could handle this conversion. And you're right, just using EGADSlite and eliminating the dependency on opencascade would make this code much more portable.

An idea... instead of making a separate pre-processing utility, would it be possible for ESP to write the EGADSLite file and the EGADS file (to support future modification of it in ESP) ?

tuckerbabcock:
I think a natural place to dump the EGADSlite model might be the preprocessing utility we already have that generates the PUMI mesh, the model file, and the supplementary model file. If we went with that route we could combine the EGADSlite file with the supplementary model into one file pretty easily too.

cwsmith:
Sounds good.

tuckerbabcock:
@cwsmith I've switched to just using EGADSlite, and no longer need the broadcast + dlopen process

tuckerbabcock:
I added back the option to build with EGADS instead of EGADSlite. In testing, I had an application that needed to use EGADS, and linking it with PUMI that used EGADSlite caused errors since EGADS and EGADSlite export the same symbols. Building the library defaults to using EGADSlite, but that can be adjusted with the option `-DUSE_EGADSLITE=OFF`.

cwsmith:
@tuckerbabcock Thanks for the update.  I think the petsc folks hit this problem so they created the following variant of EGADSlite with different function prefixes ("Original coding resulted in conflicts due to functions having the same name & signature"):
https://github.com/bldenton/EGADSlite
Would using their modified APIs make sense?  There seems to be a bit of a risk to not using the 'official' MIT version...

tuckerbabcock:
My personal inclination is to stick with the "official" release, since it would allow better support should we need it. My other thought is that if one part of an executable needs to use EGADS for some reason, I would think that having PUMI also then revert to EGADS would not dramatically reduce performance (since whatever needed EGADS is likely to be the slowest part already), though is an untested assumption.

cwsmith:
That is fine with me.  What is the performance concern with EGADS enabled vs EGADSlite enabled?  Is it the broadcast of model info?

Jumping back to this comment: https://github.com/SCOREC/core/pull/330#issuecomment-808490780
I'd suggest having all ranks either use EGADS or EGADSlite.  Hopefully this avoids the dlopen complexities.  Were you already heading down this route with the latest change?

Note, the flag `-DUSE_EGADSLITE` should be prefixed with `PUMI_` to avoid collisions with other libraries.

tuckerbabcock:
> What is the performance concern with EGADS enabled vs EGADSlite enabled? Is it the broadcast of model info?

My comment was getting at that if an application needed EGADS it would need it because somewhere it would do more than just query the geometry (it would either creating or modifying it, which would be more expensive than a query). I was also thinking about the EGADSlite paper I linked above that shows that the EGADSlite functions are thread safe and scalable, and are in general much faster than the EGADS alternatives (Table 2 from that paper).

> I'd suggest having all ranks either use EGADS or EGADSlite. Hopefully this avoids the dlopen complexities. Were you already heading down this route with the latest change?

Yes, that is the current state of the code. Based on the flag `-DUSE_EGADSLITE` (which I'll update to have the `PUMI_` prefix), it will either compile against EGADS or EGADSlite.

CMakeLists.txt
```diff
@@ -92,6 +92,14 @@ set(Trilinos_PREFIX "" CACHE STRING "Trilinos installation directory")
 option(SKIP_SIMMETRIX_VERSION_CHECK "enable at your own risk; it may result in undefined behavior" OFF)
 option(ENABLE_SIMMETRIX "Build with Simmetrix support" OFF)
 message(STATUS "ENABLE_SIMMETRIX: ${ENABLE_SIMMETRIX}")
+option(ENABLE_EGADS "Build with EGADS support" OFF)
+message(STATUS "ENABLE_EGADS: ${ENABLE_EGADS}")
+if(ENABLE_EGADS)
+  add_definitions(-DHAVE_EGADS)
```
jacobmerson, line 98: To add compiler definitions you should be [`target_compile_definitions`](https://cmake.org/cmake/help/latest/command/target_compile_definitions.html). Since you use this option in the header it will need to be public. Because this leaks into the user's downstream code you should use a prefixed name. I.e. `PUMI_HAVE_EGADS`

tuckerbabcock: Thanks for taking another look at this @jacobmerson. I wrote the `-DHAVE_EGADS` definition to be consistent with the simmetrix definition:  
https://github.com/SCOREC/core/blob/30a332d97469e9719313e312512b2d407e912f54/CMakeLists.txt#L97-L99
Based on the way that `HAVE_SIMMETRIX` is used, it seems convenient to use `add_definitions` so that it applies to all the executables so that multiple calls to `target_compile_definitions` can be skipped, but I am aware that this approach is considered bad practice in modern cmake. In any case I thought it would be best to stay consistent, though I'm not tied to the current approach.

jacobmerson: I would prefer that all of those other `HAVE_XXX` definitions get prefixed eventually and you have a valid point about consistency. The problem with adding compile definitions without a namespace is that any other consuming project linking against `pumi_egads` has that compile definition on their target. So, what if a consumer wants to use PUMI compiled with egads but they have conditional code behind `HAVE_EGADS` that they don't want to compile for some reason...currently they cannot do that since compile definitions are infectious. I think @cwsmith has last word on this sort of thing since he has to deal with all the xsdk people/rules.

The issue with the non-target version of `add_definitions` (especially in the top level cmake file) is that it will be applied to every single target and library at the current or lower sub-directory level. So, even random libraries that have nothing to do with egads are getting that compile definition added. Since these compile definitions will be infectious to anyone who uses any SCOREC target it has the potential to cause issues for down stream users.

tuckerbabcock: @cwsmith do you have any thoughts on the concerns raised by @jacobmerson?

cwsmith: @jacobmerson is correct.  We should have all the `*HAVE*`,`*ENABLE*`, etc... variables prefixed with `PUMI_` to avoid conflicts with packages that use PUMI.  +1 for using `target_compile_definitions` as well.

I created an issue to clean this up in `develop` after this is merged. https://github.com/SCOREC/core/issues/337

cwsmith (commented):
Thank you @tuckerbabcock.  I'm glad to see that this is still active.  It is a good addition to PUMI and I'd like to merge it soon.

Would you mind writing up something on the https://github.com/SCOREC/core/wiki/Mesh-Generation wiki page to discuss, or provide a reference to, how EGADS models and meshes can be used?

Besides the CMake changes discussed above, are there any other changes you intend to make?  If not, I think we should merge this and resolve the CMake changes within the scope of a larger PUMI CMake refactor.  @jacobmerson Does that make sense to you?

cmake/FindESP.cmake
```diff
@@ -0,0 +1,256 @@
+#[=======================================================================[.rst:
+FindESP
+---------
```
jacobmerson, line 3: @tuckerbabcock is this something you wrote? The style of the commenting looks different which indicates it might come from elsewhere. If so, we need to include the source/licensing information.

jacobmerson:
@cwsmith which cmake changes are you thinking we should push off vs include with this pull request?

cwsmith:
> @cwsmith which cmake changes are you thinking we should push off vs include with this pull request?

I think we can skip the changes to variable names to include the PUMI prefix for now.  The `target_compile_definitions` change is worth making now.

jehicken:
Hi @cwsmith: did you mean to assign this issue to me?  Did you mean @tuckerbabcock?  Tucker is our expert on all things EGADS.

cwsmith:
> Hi @cwsmith: did you mean to assign this issue to me? Did you mean @tuckerbabcock? Tucker is our expert on all things EGADS.

@jehicken Not intentionally.  Sorry about that.

tuckerbabcock:
Thank you @cwsmith for the renewed interest in this PR.

> Would you mind writing up something on the https://github.com/SCOREC/core/wiki/Mesh-Generation wiki page to discuss, or provide a reference to, how EGADS models and meshes can be used?

I'd be happy to write something like that up, but right now the mesh generation stuff isn't really polished at all. You have to use the ESP "analysis interface module" (AIM) I wrote for PUMI, with a slightly modified version of the ESP distribution (which I've been working on here: https://github.com/tuckerbabcock/EngSketchPad). Once you've got your mesh, you can use the gmi_egads interface I've written here with any ESP distribution. Long term, the mesh generation capabilities should not require using the modified ESP distribution (the AIM should be able to exist as a stand-alone plug-in), but I've set it up the way it is currently to make my life a little easier. 

> Besides the CMake changes discussed above, are there any other changes you intend to make? If not, I think we should merge this and resolve the CMake changes within the scope of a larger PUMI CMake refactor. @jacobmerson Does that make sense to you?

No, I think the gmi_egads interface is mostly feature complete as is right now (for everything I've needed to use it for anyway). Of course I occasionally find bugs while working on it, but those can easily be submitted as their own PRs in the future.

gmi_egads/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+  return()
+endif()
+
+if(NOT ENABLE_EGADS)
+  return()
+endif()
+
+# this file brings PUMI_USE_EGADSLITE from CMake to C++
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/gmi_egads_config.h.in"
+               "${CMAKE_CURRENT_BINARY_DIR}/gmi_egads_config.h")
```
tuckerbabcock, line 12: @jacobmerson this config file only defines `PUMI_USE_EGADSLITE`. Is conditionally calling `target_compile_definitions` a better approach here? It seems like it would be to me.

gmi_egads/pkg_tribits.cmake
```diff
@@ -0,0 +1,23 @@
...
+set(SOURCES gmi_egads.c)
+set(HEADERS gmi_egads.h)
+
+#Library
+tribits_add_library(
+  gmi_egads
+  HEADERS ${HEADERS}
+  SOURCES ${SOURCES})
+
+tribits_package_postprocess()
```
tuckerbabcock, line 23: I definitely don't understand what tribits is and I think I mostly copied this file from a different one in PUMI, so I don't have a clue if this is wrong.

mds/mdsUgrid.cc
```diff
@@ -273,6 +275,7 @@ namespace {
         verts[j] = lookupVert(r, vtx[id*nverts+j]);
         apf::Vector3 vtx_coord;
         r->mesh->getPoint(verts[j], 0, vtx_coord);
+        std::cout << "";
```
tuckerbabcock, line 278: I'll delete this.

gmi_egads/CMakeLists.txt
cwsmith: `target_compile_definitions` would be used instead of https://github.com/tuckerbabcock/core/blob/afb21d570ad03e75c346b4a45b714a0be1dbb815/CMakeLists.txt#L103

An example of its use is here:

https://github.com/SCOREC/core/blob/58a8832c9b8306fc62312ff8234479a05eeecba0/CMakeLists.txt#L159

tuckerbabcock: @cwsmith would you prefer the target be the interface target `core`, or just the `gmi_egads` target? I think the `gmi_egads` target would be more explicit and therefore preferable, but I don't have strong thoughts. 

For my above thought on having `PUMI_USE_EGADSLITE` be set with `target_compile_definitions`, I was imagining it would be set as a `PRIVATE` property of `gmi_egads`, since I think that would be semantically equivalent to the current setup with the config file.

gmi_egads/pkg_tribits.cmake
cwsmith: [Tribits](https://tribits.org/) is the cmake extension used by Trilinos.  When we add a new code this wrapper is needed.

gmi_egads/CMakeLists.txt
cwsmith: I spent some more time looking at the code and have a hopefully more complete response.  

IIUC, the user needs to know if PUMI was installed with EGADS and/or EGADSLite support and `gmi_egads/gmi_egads.c` needs to know if EGADSLite is enabled (`gmi_egads.c` is only compiled if egads is enabled).  If that is correct, I think the best option is to use `target_compile_definitions(...)` for both variables and remove the config header.  We would also want to replace any use of `HAVE_EGADS[LITE]` with the `PUMI_HAS_EGADS[LITE]` variable (or equivalent).  

To answer your questions:

- target:  Try setting it on `gmi_egads`, but that may cause problems compiling tests that use the preprocessor variables when gmi_egads is not available/installed.
- visibility: If the first assumption was correct, then I think they should both have `INTERFACE` scoping/visibility.

gmi_egads/pkg_tribits.cmake
tuckerbabcock: @cwsmith how does Tribits interact with `target_compile_definitions` set in the regular CMakeLists.txt file? Should they be the same in this file?

apf/apfMesh.cc
```diff
@@ -196,6 +196,16 @@ void Mesh::getParamOn(ModelEntity* g, MeshEntity* e, Vector3& p)
...
+
+  // above comparison only compares pointer address, this check's the entity's
+  // dim and tag in case it's the same entity but a different point to it 
+  int from_dim = getModelType(from_g);
+  int from_tag = getModelTag(from_g);
+  int to_dim = getModelType(g);
+  int to_tag = getModelTag(g);
+  if ((from_dim == to_dim) && (from_tag == to_tag))
+    return getParam(e, p);
+
```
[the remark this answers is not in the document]
tuckerbabcock: I removed this.

gmi_egads/CMakeLists.txt
tuckerbabcock: @cwsmith I've switched over all the EGADS stuff to use `target_compile_definitions`, and renamed the flags from `HAVE_EGADS` to `PUMI_HAS_EGADS`. `PUMI_HAS_EGADS` is a private definition on `gmi_egads` since it is only needed in `gmi_egads.c`, while `PUMI_HAS_EGADS` is an interface definition on `gmi_egads` since its needed by people who use the library. I did not experience any issues compiling and running the tests that use the these variables when gmi_egads wasn't built, so I think it should be okay.

jacobmerson:
@tuckerbabcock do you have instructions for how to install/run the other necessary components?

Is the only test of this functionality currently the "split" test?

@cwsmith W.R.T. testing of the geometric kernels are there any canonical tests that should be performed?

tuckerbabcock:
> @tuckerbabcock do you have instructions for how to install/run the other necessary components? Does EGADS always require ESP? Unless I had read the comments on this thread I would have no idea what ESP is.

I don't have any specific instructions, other than what's in the readme for how to build ESP: https://acdl.mit.edu/ESP/. ESP (The Engineering Sketch Pad) contains EGADS among other components. One would typically use the ESP interface to build an EGADS model. EGADS is to parasolid what ESP is to NX.

> Is the only test of this functionality currently the "split" test?
>
> @cwsmith W.R.T. testing of the geometric kernels are there any canonical tests that should be performed?

I also added a "verify" test, but you're right, there should be more. @cwsmith please let me know if there is some list of canonical tests I should add.

Issue #334, opened 2021-03-17
Compiler error gcc 10.1.0


jacobmerson:
Default build using gcc 10.1.0

```
In file included from /usr/include/string.h:633,
                 from /lore/mersoj/boundary-conditions-deps/core/phasta/phIO.c:2:
/lore/mersoj/boundary-conditions-deps/core/phasta/phIO.c: In function 'find_header':
/lore/mersoj/boundary-conditions-deps/core/phasta/phIO.c:84:7: error: '__builtin_strncpy' specified bound depends on the length of the source argument [-Werror=stringop-overflow=]
   84 |       strncpy(found, hname, strlen(hname));
      |       ^~~~~~~
/lore/mersoj/boundary-conditions-deps/core/phasta/phIO.c:84:7: note: length computed here
```

cwsmith:
I'm not sure why, but some of the phasta pre-processor (`chef`) tests failed.  I reverted the commit (https://github.com/SCOREC/core/commit/897ffdad5afb65194d769ef3d290a31eb6987a70) until we can test/investigate this some more.

cwsmith:
This is fixed again with https://github.com/SCOREC/core/commit/06119fc71d6599fbab531c24b20370b08aa1465a.  I didn't see any chef errors running on SCOREC Rhel7 and my arch linux desktop.

cwsmith:
Spoke too soon.  The errors appear in tests using Simmetrix: https://my.cdash.org/test/34314082

cwsmith:
`chef7` 
-  is a diff of the output from `chef6` against 'good' output
- `chef6` runs without any obvious issues
- Given the change, it is likely a string in the gcc10 build's output that changed.


`ph_adapt`
- loads a simmetrix geomsim smd model and scorec smb mesh 
- reads from `1-procs_case` directory of the test directory
- In the gcc 10 build the fields attach correctly, and there is a failure while accessing/checking the `motion_coords` field.  The full output is below.  This appears to be a completely different error mode than seen in the nightly build+test using gcc 7.

```
Start testing: Jun 11 16:06 EDT                                                                                                                                                                                                                                                 
----------------------------------------------------------                                                                                                                                                                                                                      
31/138 Testing: ph_adapt                                                                                                                                                                                                                                                        
31/138 Test: ph_adapt                                                                                                                                                                                                                                                           
[...]

PR #335, opened 2021-03-19
fixes issue #334

jacobmerson:
fixes the compiler error associated with the strnlen on the wrong
input of strncpy.

This needs testing/inspection from someone who is familiar with the Phasta use case.


cwsmith:
@jacobmerson Thank you.  This is the correct fix as discussed (here)[https://en.cppreference.com/w/c/string/byte/strncpy]:

>  If, after copying the terminating null character from src, count is not reached, additional null characters are written to dest until the total of count characters have been written.

cwsmith:
I'm not sure why, but some of the phasta pre-processor (`chef`) tests failed.  I reverted the commit (https://github.com/SCOREC/core/commit/897ffdad5afb65194d769ef3d290a31eb6987a70) until we can test/investigate this some more.

PR #339, opened 2021-06-07
Gcc11 fixes

sebwolf-de:
Hi, I had some compilation errors with gcc 11.1.1 on Fedora 34. I fixed the compilation errors, mostly it was changing `mallinfo()` to `mallinfo2()`, which was introduced in February 21, so maybe you'll need a flag to distinguish between glibc 2.33 and older versions.


cwsmith:
Thank you @sebwolf-de.   Your commit has been merged into develop here:

https://github.com/SCOREC/core/commit/06119fc71d6599fbab531c24b20370b08aa1465a

along a few additional commits that add logic to support the older versions of glibc.

Thanks again.

PR #344, opened 2021-07-04
Hiding Adapt Options From the Users

mortezah:
This is achieved by making the return value of `configure` APIs const. This way normal users cannot change the options after the call to `configure`. Advanced users can get a modifiable `Input` pointer by calling `ma::makeAdvanced()` API and then modify the options manually, if they choose to. All the `adapt` APIs will work with both `const Input*` and `Input*` objects.

Todo:

- [x] clean up of test (i.e., some of them really don't need to change the input)


ma/ma.cc
```diff
@@ -115,6 +120,11 @@ void adaptVerbose(Input* in, bool verbose)
   apf::printStats(m);
 }
 
+void adaptVerbose(const Input* in, bool verbose)
+{
+  adaptVerbose(makeAdvanced(in), verbose);
+}
+
```
jacobmerson, line 127: make `Adapt` take `const Input*` and this function becomes unnecessary.

jacobmerson (commented):
I think this is a big step in the right direction. I think two major issues still need to be fixed:
1. using `const_cast` in the current way causes undefined behavior whenever someone changes a setting with the  non-const object that they get back. `make_advanced` needs to make a copy instead
2. `Adapt` should take a `const Input*` object since it doesn't change anything in the `Input` object. This also avoids the need to call `createAdvanced` every time you want to adapt the solution with default settings.

ma/maLayerCoarsen.cc
```diff
@@ -190,7 +190,7 @@ static void migrateForLayerCollapse(Adapt* a, int d, int round)
 }
 
 void localizeLayerStacks(Mesh* m) {
-  Input* in = configureIdentity(m);
+  Input* in = makeAdvanced(configureIdentity(m));
   Adapt* a = new Adapt(in);
```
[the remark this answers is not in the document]
mortezah: @jacobmerson mesh adapt changes the input internally. So I will need both versions of these functions.

ma/ma.cc
```diff
@@ -51,6 +51,11 @@ void adapt(Input* in)
   apf::printStats(m);
 }
 
+void adapt(const Input* in)
+{
+  adapt(makeAdvanced(in));
+}
+
```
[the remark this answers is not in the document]
mortezah: @jacobmerson mesh adapt changes the input internally. So I will need both versions of these functions.

ma/ma.cc
mortezah: @jacobmerson mesh adapt changes the input internally. So I will need both versions of these functions.

ma/maInput.cc
```diff
@@ -285,7 +285,12 @@ Input* configureIdentity(Mesh* m, SizeField* f, SolutionTransfer* s)
   in->maximumIterations = 0;
   in->shouldFixShape = false;
   in->shouldSnap = false;
-  return in;
+  return (const Input*)in;
+}
+
+Input* makeAdvanced(const Input* in)
```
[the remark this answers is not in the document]
mortezah: @jacobmerson I haven't seen an undefined behavior, using this approach. In any event, I will implement the `clone` just to be safe.

ma/maInput.cc
```diff
@@ -213,32 +213,32 @@ Input* configure(
...
+const Input* configure(
     Mesh* m,
     IsotropicFunction* f,
     SolutionTransfer* s)
 {
   Input* in = configure(m,s);
   in->sizeField = makeSizeField(m, f);
   updateMaxIterBasedOnSize(m, in);
-  return in;
+  return (const Input*)in;
```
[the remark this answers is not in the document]
mortezah: fixed in https://github.com/SCOREC/core/pull/344/commits/11f56878f1264252c0cc7f7053dc306165fc180b

ma/maInput.cc
```diff
@@ -248,28 +248,28 @@ Input* configure(
   Input* in = configure(m,s);
   in->sizeField = makeSizeField(m, sizes, frames, logInterpolation);
   updateMaxIterBasedOnSize(m, in);
-  return in;
+  return (const Input*)in;
```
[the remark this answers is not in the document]
mortezah: fixed in https://github.com/SCOREC/core/pull/344/commits/11f56878f1264252c0cc7f7053dc306165fc180b

ma/maInput.cc
```diff
@@ -285,7 +285,12 @@ Input* configureIdentity(Mesh* m, SizeField* f, SolutionTransfer* s)
   in->maximumIterations = 0;
   in->shouldFixShape = false;
   in->shouldSnap = false;
-  return in;
+  return (const Input*)in;
```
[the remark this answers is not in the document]
mortezah: fixed in https://github.com/SCOREC/core/pull/344/commits/11f56878f1264252c0cc7f7053dc306165fc180b

ma/maInput.cc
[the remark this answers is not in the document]
mortezah: added a copy c-tor here  https://github.com/SCOREC/core/pull/344/commits/11f56878f1264252c0cc7f7053dc306165fc180b and updated `makeAdvanced` to use that instead of const_cast

jacobmerson:
@cwsmith can we rely on having c++11? Really what we want here is a move. The move constructor could just swap the pointers and then the deletion issue that @mortezah mentioned is no longer an issue.

cwsmith:
@jacobmerson Sure.  I'm fine with requiring C++11.  If we do go that route, can we make a separate PR for the changes?

jacobmerson:
@mortezah are you comfortable with move semantics? If not I can write up the prototype.

mortezah:
@jacobmerson if I understand all I need to do is to have a default move constror in Input class. Then in makeAdvance I can just return std::move(in). Is that what you where thinking?

@cwsmith why do we need a new PR?

cwsmith:
@mortezah I'd prefer not to embed the c++ standard requirement change in a series of commits related to meshadapt.  Also, I suspect we can enable the c++11 requirement with a few edits in our CMake files, quickly test with a few downstream applications (mfem, etc.), then merge to develop/master.

mortezah:
> @mortezah I'd prefer not to embed the c++ standard requirement change in a series of commits related to meshadapt.  Also, I suspect we can enable the c++11 requirement with a few edits in our CMake files, quickly test with a few downstream applications (mfem, etc.), then merge to develop/master.

So you want this branch be merged without using move semantics. And then another branch/PR where we add the move semantics?

cwsmith:
We'd merge develop/master back into this branch so we can use the c++11 features.

mortezah:
> We'd merge develop/master back into this branch so we can use the c++11 features.
> 
> I could make the c++11 change tonight if you guys are almost done with this work.

@cwsmith Yes. I think this is done. Once the c++ 11 features are in place on the develop branch, let me know and I will pull them here and use std::move in makeAdvanced.

cwsmith:
OK. I'll work on this now.

cwsmith:
@mortezah @jacobmerson the Cxx11 pr has been merged into `develop`.  `develop` can be merged into this branch.

cwsmith:
@mortezah When zoltan is disabled the meshadapt ctests 

`aniso_ma_serial`, 
`aniso_ma_serial_log_interpolation`, and 
`ma_serial`

fail with the following output:

```
MeshAdapt input error:
core is not compiled with Zoltan. Use a different balancer or compile core with ENABLE_ZOLTAN=ON!
```

(detailed log https://github.com/SCOREC/core/pull/347/checks?check_run_id=3053420768#step:8:86)

`aniso_ma_test.cc` is the source for the two `aniso_ma_serial*` tests and requires pre-zoltan balancing:

https://github.com/SCOREC/core/blob/d828f504329a55fa05b01e6f9874cf4737d2e088/test/aniso_ma_test.cc#L57

`ma_test.cc` is the source for `ma_serial` and requires pre-zoltan here:

https://github.com/SCOREC/core/blob/d828f504329a55fa05b01e6f9874cf4737d2e088/test/ma_test.cc#L58

This PR https://github.com/mortezah/core/pull/3 adds/moves the `ENABLE_ZOLTAN` conditional to prevent these tests from running when zoltan is disabled.  

Note, since some of tests are serial we could check if the process count was one and then disable pre-zoltan balancing.  Given that these are not utilities we would expect a user to run it seemed like it was not worth the added complexity.

mortezah:
@cwsmith I think all those tests can be run with parma without any problem. I will have to double check that. If that's the case I will remove the run pre Zoltan from them during the cleanup of this PR. Otherwise I will pull in the changes in the PR you mentioned above.

mortezah:
> @cwsmith can we rely on having c++11? Really what we want here is a move. The move constructor could just swap the pointers and then the deletion issue that @mortezah mentioned is no longer an issue.

@jacobmerson  I don't seem to be able to get this to work with `std::move`. I have tried a few things:

1. moving the input to `makeAdvanced` (which is of type `const ma::Input*` and returning it. Compiler error of not being able to move a const to a non-const.
2. Having the two pointer `sizeField` and `solutionTransfer` be moved inside the constructor using the `std::move` instead of just assigning the pointers. That gives me the same problem as before when I delete the original input in `makeAdvanced`.

Am I missing anything, here?

ma/maInput.cc
```diff
@@ -26,6 +26,45 @@ Input::~Input()
     delete solutionTransfer;
 }
 
+Input::Input(Input&& in)
+{
+  mesh = in.mesh;
+  sizeField = std::move(in.sizeField);
```
jacobmerson, line 32: move doesn't actually move anything...just casts to r-value reference, so this isn't doing what you expect. This should just be `std::swap`. Which will put the implicitly initialized `nullptr` to the old input.

ma/maInput.cc
```diff
@@ -26,6 +26,45 @@ Input::~Input()
...
+  ownsSizeField = in.ownsSizeField;
+  solutionTransfer = std::move(in.solutionTransfer);
```
jacobmerson, line 34: `std::swap`

ma/maInput.h
```diff
@@ -35,6 +35,9 @@ class Input
 {
   public:
     ~Input();
+    Input() {} // default empty c-tor
+    Input(const Input& in) = default; // copy c-tor
```
jacobmerson, line 39: Unless there's a good reason to keep it let's get rid of this copy since we have the move constructor which has more clear ownership semantics.

ma/maInput.h
```diff
@@ -165,21 +168,25 @@ Input* configure(
...
 /** \brief generate a no-op configuration */
-Input* configureIdentity(Mesh* m, SizeField* f=0, SolutionTransfer* s=0);
+const Input* configureIdentity(Mesh* m, SizeField* f=0, SolutionTransfer* s=0);
 
 /** \brief used internally, but users can call this if they want */
 void validateInput(Input* in);
 
+/** \brief removes constant-ness from a constant Input pointer
+           so users can modify it */
+Input* makeAdvanced(const Input* in);
```
jacobmerson, line 188: Since we are stuck using pointers for Input the semantics is unclear. This requires a comment that original the input gets put into a invalid state after this function is called.

ma/maInput.h
mortezah: @jacobmerson 

I am either missing something or this cannot be achieved by move semantics. It seems that this line

```
return new Input(std::move(*in));
```

in the current implementations of `makeAdvanced` actually uses the copy c-tor and not the move one (my best guess is because `in` is a constant). If I remove the copy c-tor I get a compile error 

```
error: use of deleted function ‘constexpr ma::Input::Input(const ma::Input&)’
```

mortezah:
@jacobmerson I think your suggestion of using move is not going to work here, since other classes used in `ma::Input`, namely `SizeField` and `SolutionTransfer`, do not have proper copy constructors implemented so calling `std::swap` to swap those pointers inside move constructor would not fix the issue with deleting the `in` object inside `makeAdvanced` after getting a new one without the `const` qualifier.

If there is no other input on this, I will switch to my earlier solution (in which I remove the contents of the destructor of the Input class at the end of any adapt routine). This is not a clean solution but since all of this happens inside adapt, it should not cause too many problems.

In my opinion, any other way of cleaning up this code would require much more in-depth cleanup in lots of other places.

ma/maInput.h
mortezah: reverted things back to pre-move semantic state since the move does not seem to work here.

ma/maInput.h
mortezah: updated this.

cwsmith:
@mortezah Is this ready for review?

mortezah:
> @mortezah Is this ready for review?

Yes.

cwsmith (approved):
This looks good to me.  How will we discourage users from calling `ma::makeAdvanced(...)`?

In addition to the w/zoltan and w/out zoltan tests that the github actions ran, I tested w/simmetrix + cad kernels + zoltan without any test failures. :+1: 

The applications downstream that use mesh adapt
- albany - we expect there to be api calls to mesh adapt; needs investigation @cwsmith 
- mfem - uses a fixed release, we can update the version and apis as needed
- m3dc1 - will look into this @mortezah 
- phasta - 'chef' calls to adapt were updated in this PR 
- other?

Running with address-sanitizer (zoltan on, simmetrix on using 17.0-210808dev) resulted in several tests failing.  They are not all using adapt so I suspect some are lingering bugs.  I'll take a quick look and the failures and annotate the list below.  Overall, it looks like the only leak related to this PR is in `vtxElmMixedBalance`.

```
The following tests FAILED:
         24 - chef-BL_query (Failed) - buffer overflow in mpi
         26 - rm_extrusion (Failed) - buffer overflow in mpi
         27 - cut_interface_sim (Failed) - leaks from Simmetrix API calls
         31 - ph_adapt (Failed) - leak in phasta getGCEssentialBCs
         57 - test_residual_error_estimate (Failed) - leaks from apf::getElement, getElementData, createMeshElement
         59 - collapse_2 (Failed) - leaks from Simmetrix API and apf
         62 - pipe_condense (Failed) - leaks from Simmetrix APIs  calls
         64 - vtxElmMixedBalance (Failed) - see below
         84 - ptnParma_cube (Failed) - agm and gmi leaks
        112 - parallel_meshgen (Failed) - leaks from Simmetrix APIs  calls
        113 - parallel_meshgen_surf (Failed) - leaks from Simmetrix APIs  calls (assumed from previous test leak)
        114 - parallel_meshgen_vol (Failed) - leaks from Simmetrix APIs  calls (assumed from previous test leak)
        115 - parallel_meshgen_para (Failed) - leaks from Simmetrix APIs  calls (assumed from previous test leak)
        120 - highOrderSolutionTransfer (Failed) - leak from iterator end not being called ?
        122 - degen_shpere_full (Failed) - leaks from Simmetrix API and apf
        123 - degen_shpere_no_north (Failed) - leaks from Simmetrix API and apf  (assumed from previous test leak)
        124 - degen_shpere_vertical_slice (Failed) - leaks from Simmetrix API and apf  (assumed from previous test leak)
        125 - crack_test (Failed) - leaks from mds calls
        126 - chefStream (Failed) - leaks from Simmetrix APIs  calls
```

`vtxElmMixedBalance` has a leak from `ma::configureIndentity`

```
                                                                                                                                                                                                                                              
=================================================================                                                                                                                                                                             
==26047==ERROR: LeakSanitizer: detected memory leaks                                                                                                                                                                                          
                                                                                                                                                                                                                                              
Direct leak of 32 byte(s) in 1 object(s) allocated from:                                                                                                                                                                                      
    #0 0x7fcd46203e77 in operator new(unsigned long) /space/cwsmith/spack/v0154_2/build_stage/spack-stage-gcc-10.1.0-tf5jjaditemasrbsl7tz6pnqa6duqwkg/spack-src/libsanitizer/asan/asan_new_delete.cpp:99                                      
    #1 0x134ffc1 in ma::setSolutionTransfer(ma::Input*, ma::SolutionTransfer*) /space/cwsmith/pumiZoltanOff/core/ma/maInput.cc:178                                                                                                            
    #2 0x134ffc1 in configure /space/cwsmith/pumiZoltanOff/core/ma/maInput.cc:190                                                                                                                                                             
                                                                                                                                                                                                                                              
Direct leak of 16 byte(s) in 1 object(s) allocated from:                                                                                                                                                                                      
    #0 0x7fcd46203e77 in operator new(unsigned long) /space/cwsmith/spack/v0154_2/build_stage/spack-stage-gcc-10.1.0-tf5jjaditemasrbsl7tz6pnqa6duqwkg/spack-src/libsanitizer/asan/asan_new_delete.cpp:99                                      
    #1 0x13513b9 in ma::configureIdentity(apf::Mesh2*, ma::SizeField*, ma::SolutionTransfer*) /space/cwsmith/pumiZoltanOff/core/ma/maInput.cc:274                                                                                             
                                                                                                                                                                                                                                              
SUMMARY: AddressSanitizer: 48 byte(s) leaked in 2 allocation(s). 
```

mortezah:
@cwsmith I will take a look at this and let you know.

cwsmith:
@mortezah Thank you.  We are all set with Albany ([albany wiki page](https://github.com/SCOREC/core/wiki/Albany-Knowledge#scorec-use-in-albany-as-of-august-2021)  The only remaining application that needs investigation is m3dc1.

mortezah:
> @mortezah Thank you. We are all set with Albany ([albany wiki page](https://github.com/SCOREC/core/wiki/Albany-Knowledge#scorec-use-in-albany-as-of-august-2021)). The only remaining application that needs investigation is m3dc1.

@cwsmith 

I will take care of m3dc. You can mark it as done here.

Also, I have committed the fix for the leak in `maLayerCoarsen.cc`. But I haven't been able to test it using address-sanitizer. That seems to require gcc10. Do we have access to that at SCOREC?

cwsmith:
@mortezah Thank you.  I marked `m3dc1` as done.

The following commands will setup an environment on SCOREC Rhel7 using gcc10:

```
# Set Environment for Rhel-7 systems, gcc 10
module unuse /opt/scorec/spack/lmod/linux-rhel7-x86_64/Core 
module use /opt/scorec/spack/v0154_2/lmod/linux-rhel7-x86_64/Core 
#the zoltan module automatically loads parmetis since it is a dependency of zoltan
module load \
gcc/10.1.0 \
mpich \
zoltan \
cmake/3.20.0  \
simmetrix-simmodsuite/17.0-210808dev \
simmetrix/simModeler/11.0-210808-dev
```

mortezah:
@cwsmith I have confirmed that my latest commit fixes the leak for `vtxElmMixedBalance`.

There are a couple of other leaks that I know how to fix. Do you want me to fix those on this PR or separately on the new one?

```
         57 - test_residual_error_estimate (Failed) - leaks from apf::getElement, getElementData, createMeshElement
        120 - highOrderSolutionTransfer (Failed) - leak from iterator end not being called ?
        125 - crack_test (Failed) - leaks from mds calls
```

cwsmith:
Sounds good.  Thank you.

I think a separate commit for the other leaks would make more sense since I assume they are not related to the changes in this PR.

cwsmith:
@mortezah I think we are ready to merge.  Are there any other changes to make here?

mortezah:
> @mortezah I think we are ready to merge. Are there any other changes to make here?

@cwsmith  I don't think so.

PR #347, opened 2021-07-12
github ci yaml

cwsmith:


cwsmith:
@jacobmerson @mortezah Travis CI stopped working a few months ago for unknown reasons (it was set it up by someone else years ago).  Earlier this year the SCOREC organization was moved to the RPI enterprise license, and, from what I understand, this gives us a lot of github 'action' minutes.  Given those changes, and that I'd prefer to avoid third-party apps when possible, this PR replaces travis CI with github actions.  

The PR builds core with GNU and Clang and minimal features (i.e., no zoltan, simmetrix, etc.) and then runs ctest on both.  The builds are triggered for pushes and pull requests to develop.  Note, nightly testing happens elsewhere before develop is merged into master.

Do you have any objections to this revised CI setup?

jacobmerson:
In theory this seems like a good idea. I don't know much about the CI landscape. The idea is that we test on whatever ubuntu's standard g++/clang++ compilers are for every pull request correct? Do we maintain a list of supported compilers? If so, do we test those nightly, or before develop->master?

cwsmith:
There is no list of list of supported compilers.  The nightly test just uses GNU compilers.

Issue #358, opened 2022-02-15
release build fails


jacobmerson:
Since `-Werror` is turned on for the release build it fails without making some modifications to the code.

As a starting place I am working on adding the release configuration to the GitHub action so it also gets tested.

jacobmerson:
Turns out it's not the release build making things fail. Maybe turning the tests off is causing issues? I'll open a new issue once I figure it out.

Issue #316, opened 2020-07-29
CMAKE_CXX_FLAGS are overwritten


jacobmerson:
The `CMAKE_CXX_FLAGS` variable is overwritten by `bob_begin_cxx_flags`. I see that the variables `SCOREC_CXX_FLAGS` and `SCOREC_EXTRA_CXX_FLAGS` can be used instead.

However, this behavior is not "standard" and I had to go into the `bob.cmake` file to figure out how to change the flags.

I think it makes more sense to have the default behavior for `CMAKE_CXX_FLAGS` do what `SCOREC_EXTRA_CXX_FLAGS` currently does.

If we don't want to go that route, then there needs to be some sort of documentation on how to set the cxx flags and a comment in the `CMakeLists.txt`

jacobmerson:
As evidence of this problem see issue #338. Symbols were not exporting due to us overwriting flags in `bob.cmake` and this made the debug process *much* harder.

KennethEJansen:
Obviously, I agree with Jacob.  I might have below average CMAKE skills but, if the intent is to control/protect the flags being set to somehow reduce mistakes that objective is not being met.  A few more "seasoned"  CMAKE gurus also had eyes on my CMAKE configuration and did not see my "foul" of adding a flag and clobbering my request for symbols.  It was not until the third person plainly told me that my binary had no symbols even though  file reported that it was not stripped that it occurred to me that adding -fpermissive to kill the warnings that were promoted to errors regarding casting MIGHT have been the culprit.

cwsmith:
The documentation here:

https://github.com/SCOREC/core/wiki/General-Build-instructions#configure

describes that `SCOREC_CXX_FLAGS` overrides while `SCOREC_CXX_EXTRA_FLAGS` adds.

Clearly, this is not documented enough, there is not enough of a warning, and/or is against cmake best practices.

I'll contact one of our contacts at Kitware about this to get their opinion.

jacobmerson:
The problem isn't so much that it's not documented, but that overriding `CMAKE_CXX_FLAGS` (or any other `CMAKE_` variable ith a `set()` command is not standard practice. Most C++ devs who have used CMake will just expect whatever you set there gets passed through.

looking through [bob.cmake](https://github.com/SCOREC/core/blob/4d659af3c5f01420013823e2a2b6a354e7d92deb/cmake/bob.cmake)

I see almost no flags that we should be explicitly setting...
[...]

PR #359, opened 2022-02-15
Add release build to github workflow

jacobmerson:
## Adding release build to GitHub actions

see issue #358 . The current release build fails, but the GitHub actions tests are passing. To make sure we notice these issues cropping up I have (attempted) to add a release build to the GitHub actions workflow.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -7,40 +7,39 @@ on:
 jobs:
   build:
     # See: https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        compiler: [g++, clang++]
+        compiler:
+          - { compiler: GNU, CC: gcc-10, CXX: g++-10 }
+          - { compiler: LLVM, CC: clang, CXX: clang++ }
+        build_type: [Debug, Release]
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        submodules: recursive
     
-    - name: install mpich
-      run: sudo apt install mpich
-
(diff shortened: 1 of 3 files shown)


jacobmerson:
It turns out the release build wasn't the only issue causing my build to fail. That being said, we should probably be testing that both the debug and release builds work since `bob.cmake` inserts a bunch of flags depending on build type and most developers don't bother to check if code is compiling with multiple build types.

jacobmerson:
To replicate the errors I was seeing I updated ubuntu to use `ubuntu-latest` which will use more modern compilers which tends to catch more issues.

Ultimately, the -Werror flag should not be in the release build as discussed in #316

cwsmith:
@jacobmerson This looks good.  Thank you.  Are you all set with this?

jacobmerson:
The pipeline works and I'm set with that. Do you need me to fix the build errors (existing issues in PUMI) before merge?

cwsmith:
Either way (new PR or here) works for me.

jacobmerson:
As I track this down I noticed that the compiler is not getting selected correctly. gcc is used for the clang build as well.

jacobmerson:
@cwsmith we are nearing completion with this and you could probably merge as is (although I should cleanup the commit history). One interesting side effect is that the `field_io` test fails for the gcc release build. This may indicate that there is some undefined behavior lurking that creates a problem with the gcc optimizer. Do you need me to try to track this down before merge?  Hopefully we can replicate with gcc-10 on the SCOREC systems.

If you are happy to merge as is, I will rebase the commits to make them nice.

cwsmith (approved):
@jacobmerson LGTM.  Thank you. 

Lets merge this and then resolve the `field_io` error in a separate PR.

jacobmerson:
@cwsmith I don't have merge permissions for core. Can you merge when you get a chance? I squashed the commits into something reasonable.

PR #372, opened 2022-09-30
Gmsh v4 off dev

KennethEJansen:
## gmshV4
Brief Description:
Changes to enable v4 gmsh file read and create a dmg that has a map from gmsh tags to unique dmg tags. In current form it breaks the existing v2 gmsh read capability which probably needs to become a separate app since v2 does not have the model info to create a dmg.

Changes were off of develop.
Addresses Issue 371.

Potential test is included below though in its current form it requires gmsh v4 or greater to be run with the imbedded info dropped into a <name>.geo file.  Alternatively, that .msh file could be generated and added to the test harness.


KennethEJansen:
Changes to enable v4 gmsh file read and create a dmg that has a map from gmsh tags to unique dmg tags.
Here is a simple channel BL example

```
// The simplest construction in Gmsh's scripting language is the 
// `affectation'. The following command defines a new variable `lc':

lc = 1e-1;
Point(1) = {0, -1, 0, lc};
Point(2) = {10, -1,  0, lc};
Point(3) = {10, 1, 0, lc};
Point(4) = {0,  1, 0, lc};

Line(1) = {1, 2}; 
Line(2) = {2, 3}; 
Line(3) = {3, 4}; 
Line(4) = {4, 1}; 

Line Loop(1) = {1, 2, 3, 4}; 
Plane Surface(1) = {1};

field1 = newf;
Field[field1]=BoundaryLayer;
Field[field1].EdgesList={1};
Field[field1].NodesList={1,2};
Field[field1].hfar=0.1;
Field[field1].hwall_n=0.001;
Field[field1].ratio=1.1;
Field[field1].thickness=0.4;
Field[field1].AnisoMax=100;
field2 = newf;
Field[field2]=BoundaryLayer;
Field[field2].EdgesList={3};
Field[field2].NodesList={3,4};
Field[field2].hfar=0.1;
Field[field2].hwall_n=0.001;
Field[field2].ratio=1.1;
Field[field2].thickness=0.4;
Field[field2].AnisoMax=100;
BoundaryLayer Field ={field1,field2};

// At this level, Gmsh knows everything to display the rectangular surface 1 and 
//+
Extrude {0, 0, 0.1} {
  Surface{1}; Layers {1}; Recombine;
}
```

If you have  v4 gmsh installed
gmsh  -3 <filename_saved_from_above>
will create a .msh file that would be used for this version of from_gmsh e.g.,
mpirun -np 1  from_gmsh <name>.dmg <name>.msh  mdsMsh/

KennethEJansen:
inf loop in verify turned to be a white space needed before line return.

Commit history also has some tries at adding various dimension of entity.  It turns out that this latest commit that adds even the zero dimension information MIGHT be working properly in PHASTA.  At least me vertex/corner BCs seem to be  being applied now.

PR #374, opened 2022-10-05
Kj/gmsh4

KennethEJansen:
## gmshV4
Brief Description:
Changes to enable v4 gmsh file read and create a dmg. In current form it breaks the existing v2 gmsh read capability which probably needs to become a separate app since v2 does not have the model info to create a dmg.

Changes were off of develop.
Addresses Issue 371.

Test   AirfoilDemo.msh file was added meshes.
`mpirun -np 1  from_gmsh airfoil.dmg AirfoilDemo.msh AirfoilDemoMDS/`
will make  `airfoil.dmg` file  and `AirfoilDemoMDS` (mds mesh directory) that are good to go with chef (assuming you make .spj file with your  BC and IC attributes).

**Note**:   gmsh native geometry kernel will make construction vertices for the center of circles that are difficult to (I could not) delete but, if you use the OpenCascade Factory to make your `.geo` file they can be deleted (see geo file for how to delete them).  To be clear they MUST be deleted or they will show up in your `dmg` file without any adjacency and `from_gmsh` will fail to pass verify.


cwsmith:
@KennethEJansen v2 support has been restored.  Did you want to sanity check that I didn't break v4 (beyond verify passing in the ctest running on the `AirfoilDemo` mesh) or should I go ahead and merge?

PR #375, opened 2022-10-11
PR for extrusion tags only

KennethEJansen:
## Extrusion Tags

Enable convert to discover the "father node" for all nodes that are extruded from one or more model faces with Simmetrix mesher.  PHASTA uses this for span wise averaging. 

Current interface is one more integer on the command line for the number of model faces and then a flat text files with the model face tags (one line per face)  is read  in a loop.

Test will be an smd, sms, xmt_txt input set and the crd, cnn, and mds file that this should create.


cwsmith:
@KennethEJansen I converted the input arg to take the file name and loop over its contents (one int per line). 
https://github.com/SCOREC/core/pull/375/commits/a15f1318e1f915142b311c48f6bc35d2c709d54f

In the input `ExtruRootID.txt` file I changed the source extrusion face to match the mesh attribute I saw in the `.smd` file:
https://github.com/SCOREC/pumi-meshes/commit/73950fdfee8547925a60dc3f456767d52a4e80ad

test/convert.cc
```diff
@@ -343,10 +343,6 @@ int main(int argc, char** argv)
   mesh->verify();
   mesh->writeNative(smb_path);
 
-  auto ft = mesh->findTag("fathers2D");
```
cwsmith: @KennethEJansen I assume removing this is OK?  The other `convert` tests don't have the "fathers2d" tag.

cwsmith (approved):

PR #380, opened 2022-12-09
matchedNodeElementReader and serial mesh support for > 2B entities

cwsmith:
## mner, convert, and > 2B node mesh support

mner, previously matchedNodeElmReader, was initially designed to take a set of text files  of connectivity, coordinates, vertex classification and optionally matching, fathers, and solution. At that time it implicitly assumed that the mesh was a box and called routine to create a `.dmg` box model.  That original version of the code was therefore limited to meshes of a topological box.  The aforementioned vertex classification used hard coded maps to relate one numbering of the `.dmg` box entities to the one created internally.  mner also previously handled multi-topology by a monolithic, rectangular connectivity with repetition of the last element node for elements having less nodes than the maximum present. 
convert, previously converted a Simmetrix `.sms` mesh to a `.smb` MDS formatted mesh.
For cases with more than 2B entities mner failed with address overflows.

In addition to addressing the >2B entities, this merge adds several features needed to support extrusions of arbitrary shape (Simmetrix->convert->MGEN->mner->chef) while preserving  MATLAB-Boxes (MATLAB->MGEN (optionally to support unstructured boxes ->mner->chef)) workflows. We drew the other workflows in because the test cases needed to work with non-rectangular representations of multi-topology connectivities added with emerging CGNS  file IO. As mner now handles arbitrary extruded models, the `.dmg` model that mner’s inputs are classified against must be created  (e.g., mdlConvert or simTranslate, or other) and is now a required input to mner. 

To make all this generality work, the vertex classification data that mner reads in was generalized to carry the tag and dimension of the model entity each mesh vertex is classified on e.g., `modelDim*1000000+dmg_model_tag`. So, 
- mesh vertices classified on **model regions** are 3000000+the model region tag number given by parasolid or by the `.dmg` file (either from mdlConvert or boxModel) ,
- mesh vertices classified on **model faces** are 2000000+the model face tag number ...,
- mesh vertices classified on **model edges** are 1000000+ the model edge tag number ..., and
- mesh vertices classified on **model vertices** are 0 + the model vertex tag number … .

convert was previously enhanced to extract the coordinates and connectivity of the extrusion root.  This allows other mesh extrusion programs to  perform the extrusion and prepare inputs for mner.  To make that process more general, in this merge, convert exports four extra columns beyond the usual x, y, coordinates in the `.crd` file.  The third column is the dimension of the model entity the vertex at that x,y, is classified on (on the root). The fourth column is that root model tag number. The fifth column is the entity (d+1) that extrudes from that root model tag. Finally, the 6th column is the model tag of the entity terminating the extrusion. This information allows new entities created by an external meshing program carrying out the extrusion to be properly classified and written to the classification file that mner requires.

@KennethEJansen, @cwsmith


phasta/phPartition.cc
```diff
@@ -37,7 +66,9 @@ apf::Migration* getSplitPlan(Input& in, apf::Mesh2* m)
       else
         splitter = apf::makeZoltanGlobalSplitter(m, method, apf::REPARTITION);
     }
-    apf::MeshTag* weights = Parma_WeighByMemory(m);
+//    apf::MeshTag* weights = Parma_WeighByMemory(m);
+    apf::MeshTag* weights = setWeights(m);
+//    Parma_PrintPtnStats(m, "preSplit",false);
```
cwsmith, line 71: Is this the logic we want?

phasta/phPartition.cc
```diff
@@ -139,9 +155,9 @@ void parmaTet(Input& in, apf::Mesh2* m, bool runGap) {
 }
 
 void parmaBalance(Input& in, apf::Mesh2* m, bool runGap) {
-  if( isMixed(m) )
-    parmaMixed(in,m);
-  else
+//  if( isMixed(m) )
+//    parmaMixed(in,m);
+//  else
```
cwsmith, line 160: I assume this should be removed?

phasta/phInput.h
```diff
@@ -23,6 +23,7 @@ class Input
   public:
     Input();
     void load(const char* filename);
+    int txtCoord; //HACK added to get through compile
```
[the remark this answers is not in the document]
cwsmith: This can likely be removed from input and output.

phasta/phPartition.cc
cwsmith: The new logic is correct.

phasta/phPartition.cc
cwsmith: Add a flag to disable the call to `parmaMixed(...)` and probably a print statement.

PR #384, opened 2023-02-24
add part count reduction to chef 

cwsmith:
## add part count reduction to chef 

- brings in controls to chef to support reducing the number of parts
- in current testing it fails at exit while freeing memory

### DIFF
--- a/phasta/chef.cc
+++ b/phasta/chef.cc
@@ -52,7 +52,12 @@ int main(int argc, char** argv)
   gmi_register_mesh();
   gmi_model* g = 0;
   apf::Mesh2* m = 0;
-  chef::cook(g,m);
+  std::string inputPath = "adapt.inp";
+  if(argc==2) inputPath = argv[1];
+  fprintf(stderr, "reading inp %s\n", inputPath.c_str());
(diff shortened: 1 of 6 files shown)


cwsmith:
@KennethEJansen 

- I rebased on develop, blew away the change to convert (please submit another pr for that), and force pushed.
- https://github.com/SCOREC/core/pull/384/commits/34c9b5888c16e736ae16c1133b2825b804f43516 resolved the problem with failing existing tests - the global objects in the anonymous namespace were causing problems
- If you have a chance would you try running the part reduction test again?

cwsmith:
there is a double free in the `Output` object created in the shrink group code here:

 https://github.com/SCOREC/core/blob/a1ae6c1c31ae7d3582a1c01ce8074d873eec99bf/phasta/phCook.cc#L255

need to track this down with valgrind.

cwsmith:
### meeting notes
- the reduced output in paraview appears OK
- @KennethEJansen will test a larger case to do another sanity check
- @KennethEJansen will look at removing the `coords.2` output file which is no longer needed

PR #385, opened 2023-03-07
A minor change to support layers in 2D mesh

usmanriaz07:
## Adds support for 2D layers in MeshAdapt.

- The support to tag elements as layers in 2D meshes has been added for the mesh adaptation. The mesh adaptation on elements (tagged as layers) can be controlled using the layer's specific input parameters.  The details of the parameters can be found in the link below:
 https://github.com/SCOREC/core/blob/master/ma/maInput.h#L119-L131

In default settings, the elements tagged as layers are not adapted at all.

- Test Case: https://github.com/SCOREC/pumi-meshes/tree/master/selectElementsAdapt
- Test Case Description:
        Elements in Original Mesh: 3264
        Elements in Adapted Mesh: 3990

The mesh elements on all model faces except one are tagged as 2d layers, and hence only elements that are not tagged as layers are adapted.


cwsmith:
Thank you for creating the PR, but it was not complete.  **Do not merge your own PR until there is an approval.**

I'm going to push the remaining needed changes to develop.

PR #387, opened 2023-03-29
Enable PCU to work without global/singleton data

jacobmerson:
## Enable PCU to work without global/singleton data

This pull request adds a new public interface in `PCU2.h` that has every PCU function take the `pcu` state as it's first argument. This enables us to use PCU in a setting where we may have PCU operating over different communicator sets.

The other modification in this pull request is to remove the `vtable` from `pcu_pmpi` as it was used sporatically and the type was assumed most places. `pcu_mpi` and `pcu_pmpi` are not part of the public interface, so users should not have been making use of this interface.

### DIFF
--- a/pcu/CMakeLists.txt
+++ b/pcu/CMakeLists.txt
@@ -28,6 +28,7 @@ set(SOURCES
 # Package headers
 set(HEADERS
   PCU.h
+  PCU2.h
   pcu_io.h
   pcu_util.h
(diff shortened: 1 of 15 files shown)


jacobmerson:
After discussion with cameron we are going to refactor PCU to be a C++ object with the appropriate state and member functions.

jacobmerson:
Succeeded by #388

PR #391, opened 2023-07-20
Fix CGNS

jrwrigh:
Removed `CGNS_ENUMV` usage for variables that didn't require it. Primary reason this worked before is that `CGNS_ENUMV` can be set such that it doesn't actually do anything, which is normally default. spack automatically enables it to prepend `CG_` to everything.

As for the other change, my compiler was giving:
```
/home/jrwrigh/software/core/mds/mdsCGNS.cc: In member function ‘void {anonymous}::MeshDataGroup::info() const’:
/home/jrwrigh/software/core/mds/mdsCGNS.cc:180:23: error: loop variable ‘m’ creates a copy from type ‘const std::pair<const int, {anonymous}::MeshData>’ [-Werror=range-loop-construct]
  180 |       for (const auto m : components)
      |                       ^
/home/jrwrigh/software/core/mds/mdsCGNS.cc:180:23: note: use reference type to prevent copying
  180 |       for (const auto m : components)
      |                       ^
      |                       &
/home/jrwrigh/software/core/mds/mdsCGNS.cc:186:23: error: loop variable ‘m’ creates a copy from type ‘const std::pair<const int, {anonymous}::MeshData>’ [-Werror=range-loop-construct]
  186 |       for (const auto m : components)
      |                       ^
/home/jrwrigh/software/core/mds/mdsCGNS.cc:186:23: note: use reference type to prevent copying
  186 |       for (const auto m : components)
      |                       ^
      |                       &
/home/jrwrigh/software/core/mds/mdsCGNS.cc:192:23: error: loop variable ‘m’ creates a copy from type ‘const std::pair<const int, {anonymous}::MeshData>’ [-Werror=range-loop-construct]
  192 |       for (const auto m : components)
      |                       ^
/home/jrwrigh/software/core/mds/mdsCGNS.cc:192:23: note: use reference type to prevent copying
  192 |       for (const auto m : components)
      |                       ^
      |                       &
cc1plus: all warnings being treated as errors
```

So I followed it's suggestion. I'll be honest, I don't know C++ super well, so I have no idea what the problem is (but kind of understand the solution).

### DIFF
--- a/apf/apfCGNS.cc
+++ b/apf/apfCGNS.cc
@@ -1051,11 +1051,11 @@ void WriteCGNS(const char *prefix, apf::Mesh *m, const apf::CGNSBCMap &cgnsBCMap
   auto communicator = PCU_Get_Comm();
   cgp_mpi_comm(communicator);
   //
-  cgp_pio_mode(CGNS_ENUMV(CGP_INDEPENDENT));
+  cgp_pio_mode(CGP_INDEPENDENT);
 
   CGNS cgns;
   cgns.fname = std::string(prefix);
-  if (cgp_open(prefix, CGNS_ENUMV(CG_MODE_WRITE), &cgns.index))
+  if (cgp_open(prefix, CG_MODE_WRITE, &cgns.index))
     cgp_error_exit();
 
   {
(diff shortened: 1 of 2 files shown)


cwsmith:
@jrwrigh These fixes were added to develop (https://github.com/SCOREC/core/commit/3e9f07b85151e80e035ad94b29d3f0d458026ed9, https://github.com/SCOREC/core/commit/4aa2326813a9214f2b1c741e0526ba4be2d86fb7, https://github.com/SCOREC/core/commit/9f91f219ccb4ba3234fd9f7c1aaf2deeb36fc4f7). Thanks for the PR; I would have spent a long time staring at the `CGNS_ENUMV`. :+1:

Issue #392, opened 2023-07-20
Build fails with PHASTA code


jrwrigh:
Trying to build core, I get:
```
In destructor ‘virtual ph::PhastaSharing::~PhastaSharing()’,
    inlined from ‘virtual ph::PhastaSharing::~PhastaSharing()’ at /home/jrwrigh/software/core/phasta/phLinks.cc:31:3,
    inlined from ‘virtual ph::PhastaSharing::~PhastaSharing()’ at /home/jrwrigh/software/core/phasta/phLinks.cc:29:12,
    inlined from ‘void ph::getLinks(apf::Mesh*, int, Links&, BCs&)’ at /home/jrwrigh/software/core/phasta/phLinks.cc:134:1:
/home/jrwrigh/software/core/phasta/phLinks.cc:29:12: error: array subscript ‘ph::PhastaSharing[0]’ is partly outside array bounds of ‘unsigned char [16]’ [-Werror=array-bounds=]
   29 |     delete helperN;
      |            ^~~~~~~
In constructor ‘ph::PhastaSharing::PhastaSharing(apf::Mesh*)’,
    inlined from ‘void ph::getLinks(apf::Mesh*, int, Links&, BCs&)’ at /home/jrwrigh/software/core/phasta/phLinks.cc:101:22:
/home/jrwrigh/software/core/phasta/phLinks.cc:24:39: note: object of size 16 allocated by ‘operator new’
   24 |     helperN = new apf::NormalSharing(m);
      |                                       ^
In destructor ‘virtual ph::PhastaSharing::~PhastaSharing()’,
    inlined from ‘virtual ph::PhastaSharing::~PhastaSharing()’ at /home/jrwrigh/software/core/phasta/phLinks.cc:31:3,
    inlined from ‘virtual ph::PhastaSharing::~PhastaSharing()’ at /home/jrwrigh/software/core/phasta/phLinks.cc:29:12,
    inlined from ‘void ph::getLinks(apf::Mesh*, int, Links&, BCs&)’ at /home/jrwrigh/software/core/phasta/phLinks.cc:134:1:
/home/jrwrigh/software/core/phasta/phLinks.cc:30:12: error: array subscript ‘ph::PhastaSharing[0]’ is partly outside array bounds of ‘unsigned char [16]’ [-Werror=array-bounds=]
   30 |     delete helperM;
      |            ^~~~~~~
In constructor ‘ph::PhastaSharing::PhastaSharing(apf::Mesh*)’,
[...]

PR #393, opened 2023-07-21
Make CGNS code C++11 compliant

jrwrigh:
This mostly consists of changing `auto` types to their actual types.

This also depends on #391 for me to be able to build with CGNS enabled locally (see the `TEMP` commit 0134d3f1fddf06cc90cd36c200d96ff278d17405).

Fixes: #392

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,9 +16,7 @@ option(USE_XSDK_DEFAULTS "enable the XDSK v0.3.0 default configuration" NO)
 #requre c++11 without extensions
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSION OFF)
-if(NOT ENABLE_CGNS)
-  set(CMAKE_CXX_STANDARD 11)
-endif()
+set(CMAKE_CXX_STANDARD 11)
 
 xsdk_begin_package()
 bob_begin_package()
@@ -27,8 +25,7 @@ if(USE_XSDK_DEFAULTS)
   xsdk_compiler_flags()
 endif()
 
-# require c++14
-option(ENABLE_CGNS "Enable the CGNS reader: requires c++14 extensions" OFF)
+option(ENABLE_CGNS "Enable the CGNS reader" OFF)
 message(STATUS "ENABLE_CGNS: ${ENABLE_CGNS}")
 
 # Set some default compiler flags that should always be used
@@ -37,10 +34,7 @@ if(NOT USE_XSDK_DEFAULTS)
   bob_begin_cxx_flags()
   bob_end_cxx_flags()
   set(CMAKE_C_FLAGS "${CMAKE_CXX_FLAGS}")
-  if(ENABLE_CGNS) #takes precedence over SCOREC_ENABLE_CXX11
-    message(STATUS "enabling cxx14")
-    bob_cxx14_flags()
-  elseif(SCOREC_ENABLE_CXX11)
+  if(SCOREC_ENABLE_CXX11)
     bob_cxx11_flags()
   endif()
 endif()
@@ -142,6 +136,14 @@ if(ENABLE_CGNS)
   set(SCOREC_USE_HDF5_DEFAULT ${ENABLE_CGNS})
   bob_public_dep(HDF5)
   add_definitions(-DHAVE_CGNS)
+else()
+  set(SCOREC_USE_CGNS_DEFAULT ${ENABLE_CGNS})
+  bob_public_dep(CGNS)
+  #CGNS does not provide cmake targets :(
+  include_directories(SYSTEM ${CGNS_INCLUDE_DIR})
+  set(SCOREC_USE_HDF5_DEFAULT ${ENABLE_CGNS})
+  bob_public_dep(HDF5)
+  add_definitions(-DHAVE_CGNS)
 endif()
 
 # Include the SCOREC project packages
@@ -172,8 +174,10 @@ add_library(core INTERFACE)
 target_link_libraries(core INTERFACE ${SCOREC_EXPORTED_TARGETS})
 if(ENABLE_CGNS)
   target_link_libraries(core INTERFACE ${CMAKE_DL_LIBS}) #HDF5 uses dlopen
-  target_compile_features(core INTERFACE cxx_std_14)
+  # target_compile_features(core INTERFACE cxx_std_14)
+  target_compile_features(core INTERFACE cxx_std_11)
 else()
+  target_link_libraries(core INTERFACE ${CMAKE_DL_LIBS}) #HDF5 uses dlopen
   target_compile_features(core INTERFACE cxx_std_11)
 endif()
 scorec_export_library(core)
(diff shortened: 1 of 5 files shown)


KennethEJansen:
I have started my work from this branch.

KennethEJansen:
@jed (that is not completing so I am not sure how Jed was on the earlier CGNS ticket but not here) contributed this picture in CNDA Slack and in the prior CGNS thread

image (11)

With the following description

- One contiguous GridCoordinates, which gives a unique global ID to each node

- TetElements followed by PyramidElements followed by PrismElements cover the volume.

- Each semantic surface (like Nacelle here) has a Elements_t node at the same level (under blk-1) with the element connectivity. This is a Pointwise convention, and it would also be possible to lump all the triangle faces into a single file node.

- Under ZoneBC, there is a BC_t node for each semantic surface, pointing to the element range (which is contiguous because of the Pointwise convention, but need not be)

- For each BC, there is also a Family_t node under Base that can offer more semantic information about that boundary type. The screenshot here doesn't use that in a meaningful way.

- Note that Symmetry contains both triangles and quads, so there are two Elements_t nodes in the file, tri_Symmetry and quad_Symmetry, with element numbering contiguous across the two. The corresponding BC_t Symmetry node refers to the contiguous range over both face topologies.

- Note that you generally don't manage writing these file nodes manually, but rather through the [Boundary Condition interfaces](https://cgns.github.io/CGNS_docs_current/midlevel/bc.html).

- To answer your question, yes this is more of the "mesh sets" model. In PETSc, we'll create the volume elements and interpolate the mesh (create the faces and edges), then broker identification of the faces through a vertex (to keep communication structured and scalable) so we can add them to the appropriate Face Sets.

KennethEJansen:
**Notes: (see CPEX 0031)
The use of ElementList and ElementRange for ptset_type is deprecated and should not be used in new code for writing boundary conditions. These are still currently accepted, but will be internally replaced with the appropriate values of PointList/PointRange and GridLocation_t, based on the base cell dimension.
Code which reads older CGNS files, should handle ElementList and ElementRange, however, since many older files contain these specifications for ptset_type.**
Is from the link @jed gave to Boundary Condition interfaces.  It looks like they don't want boundary conditions to be given in terms of face sets which is unfortunate in my opinion. Boundary conditions are attributes of model faces.  Mesh "nodes" on model edges and model vertices are best resolved by inheritance rules from the intersection of faces.  Obviously we can do that work on the SCOREC core side and produce node lists for each boundary condition but I am not clear yet whether PETSc wants these node lists or wants to do those inheritance rules and intersection from faces lists.

jedbrown:
Regarding BCs, `GridLocation_t` can take the value `FaceCenter`, so this is just a naming thing (preferring to call it `PointList` with `FaceCenter` instead of `ElementList`). Notable that Pointwise is still creating files with the old convention.

jedbrown:
> As I look at the [documentation](https://cgns.github.io/CGNS_docs_current/sids/gridflow.html#Elements), it shows ElementType_t between ElementRange and ElementConnectivity. Does cgnsview just not show this or, it it promoting that to the foldername?

You'll see it if you click on the "folder" nodes. The `10` appearing on the right is the enum value from the source code
image
```c
typedef enum {
  CGNS_ENUMV( ElementTypeNull  ) =CG_Null,
  CGNS_ENUMV( ElementTypeUserDefined ) =CG_UserDefined,
  CGNS_ENUMV( NODE ) =2,
  CGNS_ENUMV( BAR_2 ) =3,
  CGNS_ENUMV( BAR_3 ) =4,
  CGNS_ENUMV( TRI_3 ) =5,
  CGNS_ENUMV( TRI_6 ) =6,
  CGNS_ENUMV( QUAD_4 ) =7,
  CGNS_ENUMV( QUAD_8 ) =8,
  CGNS_ENUMV( QUAD_9 ) =9,
  CGNS_ENUMV( TETRA_4 ) =10,
  CGNS_ENUMV( TETRA_10 ) =11,
  CGNS_ENUMV( PYRA_5 ) =12,
  CGNS_ENUMV( PYRA_14 ) =13,
  CGNS_ENUMV( PENTA_6 ) =14,
  CGNS_ENUMV( PENTA_15 ) =15,
  CGNS_ENUMV( PENTA_18 ) =16,
  CGNS_ENUMV( HEXA_8 ) =17,
[...]
```

> What FSFBrackets are right after PrismElements and before ZoneBC?

The ordering in `cgnsview` is arbitrary from what I can tell. It's just a surface with a few hundred triangles, and has a corresponding entry in `ZoneBC`.

> Each semantic surface (like Nacelle here) has a Elements_t node at the same level (under blk-1) with the element connectivity. This is a Pointwise convention, and it would also be possible to lump all the triangle faces into a single file node. If I understand this correctly, these are a single element group for each model face following the same convention as the volume elements above. Specifically, I think the first set will range from 1 to the number of mesh faces on the first model face, followed by ElementType, followed by the connectivity table.

Mostly, though usually one writes the volume elements first so those start at `1` and the face numbers start after that. If you click around in this file, the element blocks are numbered sequentially with tet, pyramid, prism, then FSFBrackets (triangles), etc. The ordering here is arbitrary. Note that usually the mapping to file schema is handled by the library (we won't be reading/writing raw HDF5, but rather calling the CGNS interfaces that create these linked nodes). But I've certainly found it useful to explore using `cgnsview` to see how a file was intended to be used.

>    Under ZoneBC, there is a BC_t node for each semantic surface, pointing to the element range (which is contiguous because of the Pointwise convention, but need not be) Do we need these or are we going to derive all boundary conditions from model face mesh sets in the prior paragraph?

They're supposed to be present and will be written when using the API. Here's [an example](https://github.com/CGNS/CGNS/blob/develop/src/tests/write_test.c#L867-L888) from the test suite (includes multi-zone unstructured)

image

>    For each BC, there is also a Family_t node under Base that can offer more semantic information about that boundary type. The screenshot here doesn't use that in a meaningful way. Same question?

This appears optional, and is not written in the unstructured test. I haven't seen an example in which it's used for something I can recognize as important.

>    OK my largest question is whether we will or will not need ZoneBC information that I have skipped above (thinking we don't need it).

We want that because it's the way BCs are identified. It is automatically written by the interface. If for some reason we need naming heuristics instead, we could handle that, but we may as well write compliant files.

> Second question is how we plan to handle periodicity. I think Jed discussed that somewhere so I will find that and surface that in another comment.

There is `cg_conn_write`. The [docs](https://cgns.github.io/CGNS_docs_current/sids/cnct.html#GridConnectivity) (excerpt quoted below) describes how one can either match vertices to donor vertices or faces to donor faces. Note that CGNS vertices are all the nodes while PETSc vertices are just corners (of arbitrary order parametric elements). So my preference is to describe periodicity in terms of faces, which can be directly fed into the isoperiodic support. Life will be easiest if those face elements are oriented the same way on the periodic and donor surfaces.

> For Abutting or Abutting1to1 interfaces, [GridLocation](https://cgns.github.io/CGNS_docs_current/sids/build.html#GridLocation) can be either Vertex or FaceCenter. When GridLocation is set to Vertex, then PointList or PointRange refer to node indices, for both structured and unstructured grids. When GridLocation is set to FaceCenter, then PointList or PointRange refer to face elements. Face elements are indexed using different methods depending if the zone is structured or unstructured.

cwsmith:
I'm very late to the discussion here.  

Using GCC 12.3.0 I had only one compiler warning (promoted to an error) when compiling CGNS (develop @ 36bc48e) with pumi that was fixed with [this commit](https://github.com/SCOREC/core/commit/3e9f07b85151e80e035ad94b29d3f0d458026ed9). 

Unless there is a strong reason, I'd prefer to leave the CGNS code as is (with the small fix) and rely on C++14.

PR #394, opened 2023-07-21
Add compile_commands.json and .cache to .gitignore

jrwrigh:
`.cache` directory holds `clangd` cached data, while `compile_commands.json` is read by `clangd`. Neither of these should be included in the repository.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,6 @@
 build
 settings.json
+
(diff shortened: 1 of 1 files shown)


cwsmith:
I'm going to close this since it is over a year old.  Please reopen if additional discussion is needed.

PR #402, opened 2023-11-11
Initialize ma::Input::debugFolder and default to working directory

bobpaw:
## Initialize ma::Input::debugFolder and default to working directory
- Update setDefaultValues().
- Check for debugFolder in verbose ma::adaptVerbose.
- This fixes a segfault when calling ma::adaptVerbose(in, true) without setting in->debugFolder first.
- Previously, not explicitly setting debugFolder was undefined behavior
- Now, when debugFolder is not set, debug mesh VTKs are placed in the working directory.

### DIFF
--- a/ma/maDBG.cc
+++ b/ma/maDBG.cc
@@ -204,7 +204,9 @@ void dumpMeshWithFlag(ma::Adapt* a,
 
   // setup file name and write the mesh
   std::stringstream ss;
-  ss << a->input->debugFolder << "/";
+  if (a->input->debugFolder) {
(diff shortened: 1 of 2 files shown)


cwsmith (changes requested):
Thank you.  It looks like on line 207 in  1dumpMeshWithFlag(...)` the same logic appears.  Would you please fix that one too?

bobpaw:
@cwsmith, I updated that line too.

cwsmith (approved):
Thank you!

PR #403, opened 2023-11-11
Fix find_package_handle_standard_args capitalization

bobpaw:
## Fix find_package_handle_standard_args capitalization
Fix CMake warning for non-matching case of requested and found package.
Removes warnings from requesting SIMMODSUITE, ZOLTAN, and PARMETIS.
Now SimModSuite, Zoltan, and Parmetis are requested and put into the same variables.


cwsmith (approved):
Thank you!

PR #411, opened 2023-12-25
Fix outdated reel_fail call

bobpaw:
## Fix outdated reel_fail call

This fixes an outdated reel_fail call on non-Linux non-OS X systems. Presumably the macro took a `FILE*` at some point.

### DIFF
--- a/pcu/reel/reel.c
+++ b/pcu/reel/reel.c
@@ -59,6 +59,6 @@ void reel_protect(void)
 #else
 void reel_protect(void)
 {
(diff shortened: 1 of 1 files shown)


cwsmith:
@bobpaw Thank you.  What OS are you testing on?  WSL?

cwsmith (approved):

bobpaw:
@cwsmith I work primarily from an M1 Mac. After the holidays I will have access to WSL on my other workstation.

PR #412, opened 2024-01-04
Increased platform support for PCU

### LINKED ISSUE #409: Build fails on non-Linux systems: error: variable has incomplete type 'struct mallinfo'
Errors/warnings:

```
/usr/ports/math/scorec-core/work/core-2.2.8/pcu/reel/reel.c:62:13: warning: incompatible pointer types passing 'FILE *' (aka 'struct __sFILE *') to parameter of type 'const char *' [-Wincompatible-pointer-types]
  reel_fail(stderr, "reel_protect only supported on Linux and OS X");
            ^~~~~~
/usr/include/stdio.h:241:16: note: expanded from macro 'stderr'
#define stderr  __stderrp
                ^~~~~~~~~
/usr/ports/math/scorec-core/work/core-2.2.8/pcu/reel/reel.c:16:28: note: passing argument to parameter 'format' here
void reel_fail(const char* format, ...)
                           ^
1 warning generated.
--- pcu/CMakeFiles/pcu.dir/pcu_mem.c.o ---
/usr/ports/math/scorec-core/work/core-2.2.8/pcu/pcu_mem.c:55:33: warning: call to undeclared function 'mallinfo'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  struct mallinfo meminfo_now = mallinfo();
                                ^
/usr/ports/math/scorec-core/work/core-2.2.8/pcu/pcu_mem.c:55:19: error: variable has incomplete type 'struct mallinfo'
  struct mallinfo meminfo_now = mallinfo();
                  ^
/usr/ports/math/scorec-core/work/core-2.2.8/pcu/pcu_mem.c:55:10: note: forward declaration of 'struct mallinfo'
  struct mallinfo meminfo_now = mallinfo();
         ^
1 warning and 1 error generated.
```

mallinfo is a linux-specific function.

Version: 2.2.8
clang-16
FreeBSD 14.0 STABLE

bobpaw:
## Increased platform support for PCU

Add CMake check for mallctl and use it in PCU_GetMem if available (for BSD).
Add private PCU link to libexecinfo on BSD.
Enable pcu/reel/reel.c functions on BSD.
Make PUMI_HAS_MALLINFO, PUMI_HAS_MALLINFO2, PUMI_HAS_MALLCTL private compiler definitions for pcu.

Fixes #409


bobpaw:
This builds successfully on our FreeBSD VM.

CMakeLists.txt
```diff
@@ -193,11 +193,21 @@ scorec_export_library(core)
...
+  endif()
   if(PUMI_HAS_MALLINFO2)
     target_compile_definitions(core INTERFACE -DPUMI_HAS_MALLINFO2)
     target_compile_definitions(pcu PRIVATE -DPUMI_HAS_MALLINFO2)
   endif()
+  if(PUMI_HAS_MALLCTL)
+    target_compile_definitions(core INTERFACE -DPUMI_HAS_MALLCTL)
+    target_compile_definitions(pcu PRIVATE -DPUMI_HAS_MALLCTL)
+  endif()
 endif()
```
bobpaw, line 211: @cwsmith Can we move this section into pcu/CMakeLists? Also, I don't think the core INTERFACE compile definition is necessary. The definition only applies to private code and test/describe.cc but I am going to fix that file to use PCU_GetMem().

cwsmith: @bobpaw Sure.  That sounds good.

cwsmith (changes requested):
LGTM.  Thank you.  There is one small change/comment.

pcu/CMakeLists.txt
```diff
@@ -39,6 +39,12 @@ add_library(pcu ${SOURCES})
 # see: https://github.com/open-mpi/ompi/issues/5157
 target_compile_definitions(pcu PUBLIC OMPI_SKIP_MPICXX)
 
+execute_process(COMMAND uname OUTPUT_VARIABLE UNAME_STR)
```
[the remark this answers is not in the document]
bobpaw: Ah ok, thanks. I couldn't find the right variable in the `cmake -LA` output and searching "BSD" only gave me CMake>=3.25 results. I'll make that change.

CMakeLists.txt
bobpaw: Reading through `test/describe.cc`, I found a second function `get_chunk()` that returns the active memory + mmapped memory. I think that is analogous to `stats.mapped` on BSD[1][2]. But, since it requires more functionality than PCU.h provides, I think it makes the most sense to just make these compile definitions PUBLIC so that they propagate to the core target. Otherwise we would need to add another function to PCU.h (which might technically constitute a minor version bump).

Also, re-implementing the `get_heap()` function using `PCU_GetMem()` has hit a road bump. `PCU_GetMem()` does not indicate if the function is unavailable on the host OS. Right now, there is no `#else` clause, so it implicitly returns 0.0, which is potentially an actual memory amount. `get_heap()` returns -1.0 if the functionality is not supported.

 If you agree that making the definitions PUBLIC is the simplest thing, I'll just add the BSD (and Apple) code to `test/describe.cc`. It might be reasonable to notify users of core somehow that the `pumi_getMem()`/`PCU_GetMem()` function isn't supported, other than the compiler warning (in a separate PR).

https://github.com/SCOREC/core/blob/4cc2da9926adcca2a6b93ab0293040ac0e50049b/test/describe.cc#L21-L51
https://github.com/SCOREC/core/blob/4cc2da9926adcca2a6b93ab0293040ac0e50049b/test/describe.cc#L65-L84

1. https://www.man7.org/linux/man-pages/man3/mallinfo.3.html
2. https://man.freebsd.org/cgi/man.cgi?malloc(3)

cwsmith: @bobpaw Where is the `get_heap()` function defined/called?  I didn't find it in `develop` (`git grep -i get_heap`).

Let's remove the additional memory queries/reporting from `test/describe.cpp` and just pass the result of  `PCU_GetMem()` into `print_stats(...)`.  I'm not aware of any users who rely on them.

I think a configure time warning is the best we can easily do to warn the users about that unimplemented `PCU_GetMem()`.  I agree, that should be in another PR.

bobpaw: My bad, meant `get_peak()`. Otherwise sounds good.

cwsmith:
@bobpaw Thank you for working on this.  Please change it to  an 'active' PR when its ready for another review.

bobpaw:
Confirmed that tests pass on our BSD box. Marking ready for review.
Note: Automated testing may require the `MPIR_CVAR_ENABLE_GPU=0` environmental variable if using a gpu-enabled MPI build with no GPU.

cwsmith (approved):
LGTM.  Thank you.

PR #413, opened 2024-01-05
Fix test/capVol CMake and compiler warnings

bobpaw:
## Fix test/capVol CMake and compiler warnings

Remove partial support for -- to end command line option processing
- This did not make sense anyway since the final two arguments are never processed. Make the include_directories PRIVATE to the executable target

This PR fixes warnings that I noticed with SCOREC_CXX_WARNINGS=ON.


cwsmith (approved):

bobpaw:
Oops. I branched from develop. Very annoying that it defaults to master.

PR #425, opened 2024-04-26
PCU: Add nompi stub

bobpaw:
## Add nompi stub

Add a stub implementation of MPI in `pcu/pcu_pnompi.c`.
Remove direct calls to MPI functions.

This adds a `SCOREC_NO_MPI` build option for users running on single-process, single-machine setups that do not want to compile with MPI. Message passing is simulated in memory.


bobpaw:
@cwsmith, build is failing for this PR. I think we should add `PCU_Init` and `PCU_Finalize` facades that wrap the MPI calls and do a `PCU_Comm_Init()` and `PCU_Comm_Free()` to remove the direct dependencies. There are a lot of executables that would need to be changed and this would become a large PR. Alternatively, we could define a replacement `MPI_Init` and `MPI_Finalize` in pcu_pnompi and link to those. Or, we could do a hybrid approach and encourage people to use `PCU_Init` in the future.

jacobmerson:
@flagdanger and I are working on a very large update of PCU that should make this all easier by getting rid of the singleton. Pull request for that coming within the next week.

jacobmerson:
@bobpaw sorting out a few things to make the pull request ready to merge, but #388 the draft pull request for replacing the singleton in PCU.

Although we replaced the interface used in the test cases, this pull request is designed to work with no modification to existing codes by using a global PCU state when the old style interface is used with `PCU_Comm_Init` or it can use local state.

bobpaw:
@jacobmerson, we will plan to use your PR for the future of core, but the sponsor wants this functionality soon, so we will use this branch until yours is finished.

cwsmith:
@bobpaw Does `PCU_Comm_Init()` check if MPI was initialized?

bobpaw:
@cwsmith, after seeing the new PCU object introduced by #433, we would still like to take this build-time stubbing approach. This PR results in compilation of the entire repository without MPI dependence. Since I merged develop a few times, should we squash this PR?

CMakeLists.txt
```diff
@@ -29,6 +29,15 @@ else()
   bob_set_cxx_standard(14)
 endif()
 
+#
+# option related to MPI
+#
+option(SCOREC_NO_MPI "Use a stub MPI implementation" OFF)
```
cwsmith, line 35:
```suggestion
option(SCOREC_NO_MPI "Disable MPI support.  When enabled only serial execution is supported." OFF)
```

pcu/PCU.h
```diff
@@ -1,14 +1,34 @@
...
+#if defined(SCOREC_NO_MPI)
+#include "pcu_pnompi_types.h"
+// Remove MPI calls.
+#define MPI_Init(argc, argv) do { \
+(void) argc; \
+(void) argv; \
+} while (0)
+#define MPI_Finalize(void) 
+#ifdef __cplusplus
+extern "C"
```
cwsmith, line 13: What is the scope of these macros?  We still want to let users of PUMI run their application in parallel even if PUMI is only built for serial execution.

I think we should add a test that builds an MPI hello world app and links against a 'SCOREC_NO_MPI=on' build, and loads a serial pumi mesh etc..  This will help ensure that the symbols don't leak in the long run.

pcu/PCU_C.h
```diff
@@ -1,7 +1,21 @@
...
+#if defined(SCOREC_NO_MPI)
+#include "pcu_pnompi_types.h"
+// Remove MPI calls.
+#define MPI_Init(argc, argv) do { \
+(void) argc; \
+(void) argv; \
+} while (0)
+#define MPI_Finalize(void)
+#ifdef __cplusplus
+extern "C"
```
cwsmith, line 13: same scope question as in PCU.h

pcu/pcu_pnompi_types.h
```diff
@@ -0,0 +1,22 @@
...
+#define PCU_PNOMPI_TYPES_H
+
+typedef int MPI_Comm;
+typedef int MPI_Request;
+typedef int MPI_Datatype;
+typedef int MPI_Op;
+#define MPI_COMM_WORLD 0
+#define MPI_ANY_SOURCE -1
+#define MPI_SUM 1
+#define MPI_INT 2
```
cwsmith, line 20: same scope question as in PCU.h

test/pumi.cc
```diff
@@ -11,7 +11,6 @@
 #include <unistd.h>
 #include <iostream>
 #include <algorithm>
-#include <mpi.h>
```
cwsmith, line 14: If SCOREC_NO_MPI=ON, then unless the user is compiling with the mpi wrappers, then source code and tests cannot `#include <mpi.h>`.  Is that correct?  If so, we should document that in the users and developers guide and that no MPI functions should be called directly within PUMI source.

test/testing.cmake
```diff
@@ -2,12 +2,36 @@ set(MESHES ""
     CACHE STRING
     "Extracted http://scorec.rpi.edu/pumi/pumi_test_meshes.tar.gz")
 function(mpi_test TESTNAME PROCS EXE)
+  if(SCOREC_NO_MPI)
+  if(${PROCS} EQUAL "1")
```
cwsmith, line 6:
```suggestion
if(SCOREC_NO_MPI AND ${PROCS} EQUAL "1")
```
Will that work?

test/smokeTesting.cmake
```diff
@@ -1,8 +1,16 @@
 function(smoke_test TESTNAME PROCS EXE)
   set(tname smoke_test_${TESTNAME})
+  if(SCOREC_NO_MPI)
+  if(PROCS EQUAL "1")
```
cwsmith, line 4:
```suggestion
if(SCOREC_NO_MPI AND ${PROCS} EQUAL "1")
```
Will that work?

cwsmith (changes requested):
LGTM.  Thank you.  There are a few comments below.

test/testing.cmake
bobpaw: No, and I updated the indentation to make it clearer what's going on. If SCOREC_NO_MPI is given, I want to only add serial tests. If AND is used instead, then parallel tests will still be added as usual.

pcu/PCU.h
bobpaw: This is a really good point. Right now, those macros would spread to every source that includes PCU.h. I think a simple `#ifndef MPI_COMM_WORLD` may suffice in these cases. Then PUMI tests will use the stub MPI_Init and end-users can `#include <mpi.h>` before PCU.h and they will get the real versions.

I will work on making a test case so that we can observe the negative first before trying that fix.

Also, it would require a major notice to end-users (since include order is not usually important). What do you think?

cwsmith: If building with SCOREC_NO_MPI=off then I assume that everything that worked prior to this PR would continue to work.  Is that correct?  If so, then I think we could count this as a new 'feature' with specific requirements (i.e., include order, no mpi headers, etc.).

bobpaw: Also, I see the Developer Guide on the Wiki which I will update after merging this PR. Is the user guide LaTeX anywhere? A quick Ctrl+F doesn't show any reference to any of the `SCOREC_*` build options.

cwsmith: > Should this be a GitHub action, similar to building the doc/ folder?

Yes.

> latex 

The latex files for the user guide are here:

https://github.com/SCOREC/pumi-docs/tree/master/userGuide

bobpaw:
Before I make my example project I have a feeling I will need to add a config file, called something like `mpi_rpc.h` (@cwsmith I would appreciate input on the name) which will be generated to make SCOREC_NO_MPI persistent. Then I can bundle the PCU.h `#ifdef SCOREC_NO_MPI` logic and the content in pcu_pnompi_types.h into one file.

cwsmith:
We should probably just make a PUMI_config.h file ([like this](https://github.com/SCOREC/meshFields/blob/main/src/MeshField_Config.hpp.in)) and the first entry in it can be the `SCOREC_NO_MPI` flag.  Over time, we can move away from using `add_definition(...)` to the config file approach.

phasta/CMakeLists.txt
```diff
@@ -102,7 +102,6 @@ endif()
 # Include directories
 target_include_directories(ph INTERFACE
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
-    $<INSTALL_INTERFACE:include>
```
jacobmerson, line 105: why is `include` not needed?

pcu/PCU.cc
```diff
@@ -207,7 +225,14 @@ void PCU::DebugPrint(const char *format, va_list args) noexcept {
   vfprintf(msg_->file, format, args);
   fflush(msg_->file);
 }
-PCU::PCU(MPI_Comm comm) {
+PCU::PCU() :
+#ifndef SCOREC_NO_MPI
+PCU(MPI_COMM_WORLD)
```
jacobmerson, line 230: You should never actually use `MPI_COMM_WORLD`, instead call `MPI_Comm_dup` on the world communicator.

pcu/pcu_pmpi.c
```diff
@@ -23,12 +23,23 @@ void pcu_pmpi_init(MPI_Comm comm, pcu_mpi_t* self)
...
+  self->owned = 0;
 }
 
 void pcu_pmpi_finalize(pcu_mpi_t* self)
 {
   MPI_Comm_free(&(self->user_comm));
   MPI_Comm_free(&(self->coll_comm));
+  // Prevent accidental freeing of MPI_COMM_WORLD.
+  int result;
+  MPI_Comm_compare(self->original_comm, MPI_COMM_WORLD, &result);
```
jacobmerson, line 35: if you use `MPI_Comm_dup` above instead of directly using `MPI_COMM_WORLD` you will not need this.

phasta/adaptLvlSet_loop.cc
```diff
@@ -36,9 +36,13 @@ namespace {
 }
 
 int main(int argc, char** argv) {
+#ifndef SCOREC_NO_MPI
   MPI_Init(&argc, &argv);
+#else
+  (void) argc, (void) argv;
+#endif
```
jacobmerson, line 43: Is there a way we can get rid of the `#ifdef` from all the user code? One potentially bad idea: 
define a `PCU_Initialize` and `PCU_finalize` functions and put this code inside those. Until we have proper MPI sessions support, I'm not sure if there is much better that we can do.

pumi/CMakeLists.txt
```diff
@@ -40,7 +40,6 @@ add_library(pumi ${SOURCES})
 # Include directories
 target_include_directories(pumi PUBLIC
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
-    $<INSTALL_INTERFACE:include>
```
jacobmerson, line 43: why is `include` not needed?

pumi/pumi_mesh.cc
```diff
@@ -437,12 +422,17 @@ void pumi_mesh_print (pMesh m, bool print_ent)
...
+    global_local_entity_count, local_entity_count,
+    sizeof(int) * 4 * m->getPCU()->Peers()
+  );
+  m->getPCU()->Add(global_local_entity_count, 4);
+
+  memcpy(
+    global_own_entity_count, own_entity_count,
+    sizeof(int) * 4 * m->getPCU()->Peers()
+  );
+  m->getPCU()->Add(global_own_entity_count, 4);
```
jacobmerson, line 435: I think this needs a comment describing what's going on here. Should there be a `PCU_Allreduce`?

pcu/PCU_C.h
bobpaw: Removed these macros and added `#ifndef SCOREC_NO_MPI` to all test cases. Additionally, the change is less obscure to someone casually reading tests as examples.

python_wrappers/CMakeLists.txt
```diff
@@ -21,7 +21,6 @@ if(ENABLE_SIMMETRIX)
   target_include_directories(simhelper PUBLIC ${SIMMODSUITE_INCLUDE_DIR})
   target_include_directories(simhelper PUBLIC
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
-    $<INSTALL_INTERFACE:include>
```
jacobmerson, line 24: why is include path removed?

ree/CMakeLists.txt
```diff
@@ -21,7 +21,6 @@ add_library(ree ${SOURCES})
 # Include directories
 target_include_directories(ree INTERFACE
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
-    $<INSTALL_INTERFACE:include>
```
jacobmerson, line 24: why is include path removed?

sam/CMakeLists.txt
```diff
@@ -17,7 +17,6 @@ add_library(sam ${SOURCES})
 # Include directories
 target_include_directories(sam INTERFACE
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
-    $<INSTALL_INTERFACE:include>
```
jacobmerson, line 20: why is include path removed?

spr/CMakeLists.txt
```diff
@@ -20,7 +20,6 @@ add_library(spr ${SOURCES})
 # Include directories
 target_include_directories(spr INTERFACE
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
-    $<INSTALL_INTERFACE:include>
```
jacobmerson, line 23: why is include path removed?

stk/CMakeLists.txt
```diff
@@ -61,7 +61,6 @@ target_link_libraries(apf_stk PUBLIC ${Shards_LIBRARIES})
 # Include directories
 target_include_directories(apf_stk INTERFACE
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
-    $<INSTALL_INTERFACE:include>
```
jacobmerson, line 64: why is include path removed?

zoltan/CMakeLists.txt
```diff
@@ -47,7 +47,6 @@ add_library(apf_zoltan ${SOURCES})
 # Include directories
 target_include_directories(apf_zoltan INTERFACE
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
-    $<INSTALL_INTERFACE:include>
```
jacobmerson, line 50: why is include path removed?

pcu/PCU.cc
```diff
@@ -232,9 +258,15 @@ PCU &PCU::operator=(PCU && other) noexcept {
   std::swap(msg_, other.msg_);
   return *this;
 }
-MPI_Comm PCU::GetMPIComm() const noexcept { return mpi_->original_comm; }
+PCU_Comm PCU::GetMPIComm() const noexcept { return GetComm(); }
+PCU_Comm PCU::GetComm() const noexcept { return mpi_->original_comm; }
```
jacobmerson, line 262: I'm wondering if we should remove this from the interface. Otherwise, it may be too tempting to drop out of `PCU` to `MPI` which will break the no mpi use case.

jacobmerson (changes requested):
I mostly focused on the changes in the C++ and CMake code and did not review the C implementation in detail.

I have one major concern that we need to clarify before merging is a more concrete understanding of the semantics of the class's comm ownership. I'm currently against a shared ownership model for this particular case. And, highly against anything that puts the onus of remembering to clean things up on the user. Is there a simple case that can demonstrate why you needed to add the `ownscomm` logic?

The other more minor concern is that we should get rid of the need for #ifdefs in the user code by providing our own `init` and `finalize` routines. That will substantially clean up the executables.

The rest is very minor stuff that should be quick to update.

Otherwise, I think this is very much moving in the right direction. The changes clean up a fair number of places where we have a mix of MPI and PCU use which is bound to cause issues.

pcu/PCU.cc
bobpaw: I think there are cases where users may want access to advanced MPI features. A simple one I can think of (I have limited MPI experience) is if a user wants to call MPI_Comm_compare.

example/mpi-nompi/hello.cc
```diff
@@ -0,0 +1,98 @@
...
+#include <gmi_mesh.h>
+#include <apf.h>
+#include <apfBox.h>
+#include <apfMesh2.h>
+
+void print_array(int *arr, size_t n);
+
+int main(int argc, char** argv) {
+  MPI_Init(&argc, &argv);
+  pcu::PCU *PCUObj = new pcu::PCU(MPI_COMM_WORLD);
```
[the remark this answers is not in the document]
bobpaw: I don't like how we have to introduce a scope just to destruct the PCUObj at a certain time. I use new and delete to explicitly destroy it before MPI_Finalize.

example/mpi-nompi/hello.cc
```diff
@@ -0,0 +1,98 @@
...
+  for (int i = 0; i < size; ++i) {
+    if (rank == i)
+      std::cout << "MPI(" << rank << ") sum: " << sum <<
+      "; PCU sum: " << psum << std::endl;
+    // Print deterministically please and thank you.
+    std::this_thread::sleep_for(std::chrono::milliseconds(5));
+    MPI_Barrier(MPI_COMM_WORLD);
+  }
+  // Test Allgather
+  int *vals1 = new int[size];
```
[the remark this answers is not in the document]
bobpaw: Raw new is not really bad in a test like this because I know I am two lines later but I have changed it to std::vector

gmi/CMakeLists.txt
```diff
@@ -31,7 +31,6 @@ add_library(gmi ${SOURCES})
 target_link_libraries(gmi pcu lion)
 target_include_directories(gmi INTERFACE
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
-    $<INSTALL_INTERFACE:include>
```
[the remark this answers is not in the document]
bobpaw: See above comment, let's discuss there.

example/mpi-nompi/hello.cc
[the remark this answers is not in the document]
bobpaw: The mpi-nompi example is an MPI application linked to SCOREC compiled with SCOREC_NO_MPI. It demonstrates and tests the combination of the two.

gmi_cap/CMakeLists.txt
```diff
@@ -18,7 +18,6 @@ target_link_libraries(gmi_cap PUBLIC gmi pcu lion capstone_module framework_test
 # Include directories
 target_include_directories(gmi_cap PUBLIC
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
-    $<INSTALL_INTERFACE:include>
```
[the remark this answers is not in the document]
bobpaw: See above comment, let's discuss there.

gmi_sim/CMakeLists.txt
```diff
@@ -26,7 +26,6 @@ target_link_libraries(gmi_sim PUBLIC gmi pcu ${SIMMODSUITE_LIBS})
 # Include directories
 target_include_directories(gmi_sim PUBLIC
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
-    $<INSTALL_INTERFACE:include>
```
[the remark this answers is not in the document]
bobpaw: See above comment, let's discuss there.

mds/CMakeLists.txt
```diff
@@ -47,7 +47,6 @@ add_library(mds ${SOURCES})
 # Include directories
 target_include_directories(mds INTERFACE
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
-    $<INSTALL_INTERFACE:include>
```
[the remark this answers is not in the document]
bobpaw: See above comment, let's discuss there.

pcu/PCU.cc
```diff
@@ -193,7 +203,15 @@ void PCU::DebugOpen() noexcept {
 
 double GetMem() noexcept { return pcu_get_mem(); }
 void Protect() noexcept { reel_protect(); }
-double Time() noexcept { return MPI_Wtime(); }
+double Time() noexcept {
```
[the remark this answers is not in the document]
bobpaw: I like that idea. In our stub we looked at the OpenMPI implementation (I think, but maybe mpich) to decide what to put. If I had to guess, they resorted to gettimeofday/other functions to ensure compatibility. But since C++ has standard clocks we don't have to screw around trying to be compatible.

pcu/PCU.cc
```diff
@@ -232,9 +258,15 @@ PCU &PCU::operator=(PCU && other) noexcept {
   std::swap(msg_, other.msg_);
   return *this;
 }
-MPI_Comm PCU::GetMPIComm() const noexcept { return mpi_->original_comm; }
+PCU_Comm PCU::GetMPIComm() const noexcept { return GetComm(); }
+PCU_Comm PCU::GetComm() const noexcept { return mpi_->original_comm; }
+bool PCU::OwnsComm() const noexcept { return mpi_->owned; }
```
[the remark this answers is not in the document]
bobpaw: `GetMPIComm` and `GetComm` are marked `[[nodiscard]]`:

https://github.com/SCOREC/core/blob/b0f5bf12e76968623e4eb637509130a3b6595db7/pcu/PCU.h#L34

Do you mean `OwnsComm(void)`? Do you want to differentiate it from `OwnsComm(bool)`?

pcu/PCU.cc
```diff
@@ -232,9 +258,15 @@ PCU &PCU::operator=(PCU && other) noexcept {
...
+void PCU::OwnsComm(bool on) const noexcept { mpi_->owned = on ? 1 : 0; }
 
-MPI_Comm PCU::SwitchMPIComm(MPI_Comm newcomm) noexcept {
+PCU_Comm PCU::SwitchMPIComm(PCU_Comm newcomm) noexcept {
```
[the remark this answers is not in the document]
bobpaw: Done

pcu/PCU.cc
```diff
@@ -232,9 +258,15 @@ PCU &PCU::operator=(PCU && other) noexcept {
...
+  return SwitchComm(newcomm);
+}
+PCU_Comm PCU::SwitchComm(PCU_Comm newcomm) noexcept {
```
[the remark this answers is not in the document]
bobpaw: Did this already and it must be in the header not the source.

pcu/PCU.cc
```diff
@@ -324,7 +361,8 @@ template <typename T> T PCU::Exscan(T p) noexcept {
   template void PCU::Max<T>(T * p, size_t n) noexcept;                         \
   template T PCU::Max<T>(T p) noexcept;                                        \
   template void PCU::Exscan<T>(T * p, size_t n) noexcept;                      \
-  template T PCU::Exscan<T>(T p) noexcept;
+  template T PCU::Exscan<T>(T p) noexcept;                                     \
```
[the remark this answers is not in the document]
bobpaw: These are explicit template instantiations, not the declaration, so not the right place for the attribute.

sam/CMakeLists.txt
bobpaw: See above comment, let's discuss there.

spr/CMakeLists.txt
bobpaw: See above comment, let's discuss there.

zoltan/CMakeLists.txt
bobpaw: See above comment, let's discuss there.

stk/CMakeLists.txt
bobpaw: See above comment, let's discuss there.

pcu/PCU.h
```diff
@@ -26,7 +26,23 @@ class PCU {
...
-  [[nodiscard]] MPI_Comm GetMPIComm() const noexcept;
+  [[deprecated("Use PCU::GetComm instead.")]]
+  [[nodiscard]] PCU_Comm GetMPIComm() const noexcept;
+  /**
+   * @brief Get the underlying communicator which may be an MPI_Comm.
+   */
+  [[nodiscard]] PCU_Comm GetComm() const noexcept;
+  /** @brief Check if the original PCU_Comm is owned by this object.
+   * If true, it will be freed during destruction. */
+  bool OwnsComm() const noexcept;
```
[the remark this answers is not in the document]
bobpaw: When `PCU::Split()` is used, a new communicator is constructed with the newcomm from MPI_Comm_split. This MPI communicator needs to be freed. Before this PR, PCU only manages `user_comm` and `coll_comm`. The `original_comm` passed to the constructor is copied and not touched during destruction (`pcu_pmpi_finalize`). We considered using something like `PCU::FreeComm` to destroy the underlying split communicator, but then the dangling `original_comm` is still accessible via `PCU::GetMPIComm`. Plus, most users will just be creating a PCU with Split and then `delete`ing or  putting it in a `std::unique_ptr` for automatic deletion. So if we have a sense of ownership and can automatically remove it, then the typical user has less to do. And, they can call `PCU::OwnsComm(false)`, extract the underlying comm with `PCU::GetComm`, and do what they want.

https://github.com/SCOREC/core/blob/437da6845c0d3c7bdc747fa10490ab547f0219a5/parma/group/parma_group.cc#L51-L64

https://docs.open-mpi.org/en/v5.0.1/man-openmpi/man3/MPI_Comm_split.3.html

bobpaw:
> The other more minor concern is that we should get rid of the need for #ifdefs in the user code by providing our own init and finalize routines. That will substantially clean up the executables.

I was considering adding a specialized `PCU::PCU(int *argc, char ***argv)` which would do MPI_Init and save a flag to run MPI_Finalize later. That specialization would allow users to avoid introducing a scope just for the PCU object because MPI_Finalize will be guaranteed to be called after any pcu_finalize, etc.

Outside of something like that, what is so bad about having #ifdefs? When we discussed, we thought it would help to make it clear what is going on to people looking at tests as examples.

pcu/PCU.cc
cwsmith: I think we should keep it.  Maybe adding a docstring with the suggestion that collective and p2p messaging that involves pumi objects should use PCU interfaces and not direct MPI calls.

cwsmith:
Are the `#ifdef SCOREC_NO_MPI` macros only used in the examples/drivers?

dsp/CMakeLists.txt
```diff
@@ -28,7 +28,6 @@ add_library(dsp ${SOURCES})
 # Include directories
 target_include_directories(apf INTERFACE
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
-    $<INSTALL_INTERFACE:include>
     )
```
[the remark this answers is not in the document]
jacobmerson: that makes sense.

bobpaw:
> Are the `#ifdef SCOREC_NO_MPI` conditionals only used in the examples/drivers?

@cwsmith, the places with `#ifdef SCOREC_NO_MPI` are: tests (including "utils"), some exes in `phasta/`, `pcu/pcu_defines.h`, and `pcu/PCU.h` (for pcu::Time and the default constructor).

pcu/PCU.cc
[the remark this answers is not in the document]
jacobmerson: good point.

jacobmerson:
The thing about ifdefs is it can make the code much more unreadable and the user needs to know under the hood whether they should have MPI or not. If we wrap those ifdefs in our own initialize function then we take care of that subtlety for them.

> I was considering adding a specialized `PCU::PCU(int *argc, char ***argv)` which would do MPI_Init and save a flag to run MPI_Finalize later. That specialization would allow users to avoid introducing a scope just for the PCU object because MPI_Finalize will be guaranteed to be called after any pcu_finalize, etc.

I don't like this solution. The reason we have the PCU object is that we may have many of them lying around with different communicators under the hood, but ultimately still using the same MPI library. I have use cases where this is needed. That constructor means the PCU object may or may not own the MPI initialization. What happens when the owning PCU goes out of scope? All other PCUs that are still alive would break as they would be doing MPI operations after `MPI_finalize`.

So, there needs to be some library/MPI initialization at a higher level than the PCU object.

example/mpi-nompi/hello.cc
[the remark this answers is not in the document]
jacobmerson: It's not a huge concern here because it's a test driver, but I prefer that we avoid new/delete in all code. Although you know what you are doing there will be grad students who are looking into these files and learning how to code by example. To me the annoyance of an additional scope is better than the bugs we get from leaky code.

pcu/PCU.cc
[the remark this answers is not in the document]
jacobmerson: This is my mistake since you marked them in the header file already.

bobpaw:
> What happens when the owning PCU goes out of scope? All other PCUs that are still alive would break as they would be doing MPI operations after MPI_finalize.

Shouldn't the world-initializing PCU only go out of scope at the end of the main function? Shouldn't all other objects have been destructed by then?

bobpaw:
I am not touching apfCGNS code yet because pcgnslib.h has `#include "mpi.h"` and therefore CGNS is incompatible with SCOREC_NO_MPI. But I will need to add CMake logic related to the incompatibility and update apfCGNS to not use GetMPIComm.

bobpaw:
We determined that instead of `PCU::OwnsComm`, we can remove `pcu_mpi_t::original_comm` entirely and also remove `PCU::GetMPIComm` and `PCU::SwitchMPIComm` since those functions only existed because we had to shuffle the MPI_Comm around in the singleton PCU. With the new `pcu::PCU` C++ object, we do not need such functions.

bobpaw:
SCOREC_NO_MPI is also likely exclusive with ENABLE_ZOLTAN since zoltan.h has `#include <mpi.h>`.

cwsmith:
> SCOREC_NO_MPI is also likely exclusive with ENABLE_ZOLTAN since zoltan.h has `#include <mpi.h>`.

Hmmm.  Zoltan can be used locally on each process (see the MPI_COMM = SELF) to do what we called 'local part splitting' but that use case then relies on migrating elements to other processes via ... MPI.  So, yes, `ENABLE_ZOLTAN` requires MPI.  Good catch.

bobpaw:
@cwsmith, I'm running into a subtle bug that I'm hoping you can help with. The test case `pumi3d-1p` is failing intermittently (1/4 or 1/5 of the time) in this new branch. I believe the culprit is a change made in pumi/pumi_mesh.cc, but I cannot figure out how my changes caused this issue. The test is only activated given `ENABLE_ZOLTAN`, and since this code uses `apf::createZoltanSplitter`, it is doing the `MPI_COMM_SELF` splitting like you described. I can additionally share logs of the failing test.

New code:

https://github.com/SCOREC/core/blob/006947468405b1e31a3a22407d10f38fa09c8236/pumi/pumi_mesh.cc#L213-L241

Old code:

https://github.com/SCOREC/core/blob/a900c4761cdc33592be0d5b0c1521cf3996a043d/pumi/pumi_mesh.cc#L234-L260

pcu/PCU.h
```diff
@@ -130,6 +125,25 @@ void Protect() noexcept;
...
+ * @brief Finalize the underlying parallel library.
+ *
+ * This may be MPI (or a stub, given SCOREC_NO_MPI). This function abstracts
+ * the difference.
+ *
+ * @return 0 on success, nonzero otherwise.
+ */
+int PCU_Finalize();
+
 PCU* PCU_GetGlobal();
```
jacobmerson, line 147: I know this isn't something you added, but can we remove this? I don't think we store a global PCU pointer anymore.

jacobmerson (approved):
Looks wonderful! Thankyou for putting this together!

pcu/PCU.h
bobpaw: Sure, it looks like there is no definition anywhere for the C++ or C api version of this call so I will remove it.

bobpaw:
@cwsmith I have made the changes you and Jacob requested. Can you review when you get the chance?

cwsmith:
@bobpaw Here is a comment that discusses running the delta wing regression test I mentioned earlier:
https://github.com/SCOREC/core/pull/433#issuecomment-2516069759

bobpaw:
> @bobpaw Here is a comment that discusses running the delta wing regression test I mentioned earlier: [#433 (comment)](https://github.com/SCOREC/core/pull/433#issuecomment-2516069759)

Do I need to worry about about timing? Or should I just compare the quality file?

cwsmith:
We should check timing and quality.

bobpaw:
> We should check timing and quality.

@cwsmith, adapt completed in 5761.497560 seconds. How do I measure the resulting length_proc.txt and quality_proc.txt files? I see there's a histogram utility but it wants min, max, nbins.

cwsmith:
@bobpaw Hmmm.  I'm guessing that was a serial run?  If so, I only have prior results for a 32 processes (see https://github.com/SCOREC/core/pull/433#issuecomment-2516069759).  Regarding quality, if the final entity counts are similar we should be OK.  We could also render the adapted mesh to sanity check.

bobpaw:
No, this was a 32proc run on wilbur.

bobpaw:
But it was a debug build without optimization so let me rerun with release optimizations.

cwsmith:
For the record, wilbur and piglet have the same hardware.

bobpaw:
Wow, I'm surprised how much optimizations improved the runtime:

MeshAdapt: mesh adapted in 308.636508 seconds
number of tet 5934317 hex 0 prism 0 pyramid 0
mesh entity counts: v 1067369 e 7026323 f 11893272 r 5934317

cwsmith:
That time looks good.  Can you please paste your cmake config command and the output log?

bobpaw:
config/deltaWing.sh:
```sh
#!/bin/sh

build_id=deltaWing
if [ $# -gt 0 ]; then
  build_id=$1
fi

cmake -S . -B build/$build_id \
  -GNinja \
  -DCMAKE_INSTALL_PREFIX="$(realpath install)/$build_id" \
  -DCMAKE_BUILD_TYPE="Release" \
  -DCMAKE_C_COMPILER="mpicc" \
  -DCMAKE_CXX_COMPILER="mpicxx" \
  -DSCOREC_CXX_SYMBOLS=OFF \
  -DSCOREC_CXX_OPTIMIZE=ON \
  -DSCOREC_CXX_WARNINGS=ON \
  -DBUILD_SHARED_LIBS=ON \
  -DBUILD_EXES=ON \
  -DIS_TESTING=ON \
  -DENABLE_ZOLTAN=ON \
  -DENABLE_SIMMETRIX=ON -DSKIP_SIMMETRIX_VERSION_CHECK=ON \
  -DENABLE_FIELDSIM=OFF -DSIM_MPI=mpich4.1.1 -DSIM_ACIS=ON -DSIM_PARASOLID=ON \
  -DPCU_COMPRESS=ON
```

<details>
<summary>
run_adapt.log
</summary>

```
model /lore/smithc11/projects/deltaWingAdapt/pumiAdapt/deltaWing.dmg loaded in 0.000793 seconds
mesh p32/ loaded in 0.062518 seconds
number of tet 581196 hex 0 prism 0 pyramid 0
mesh entity counts: v 104305 e 691152 f 1168044 r 581196
  - verifying tags: mach_ver, original_metric_ver, target_metric_ver
mesh verified in 0.148043 seconds
writeVtuFile into buffers: 0.030325 seconds
writeVtuFile buffers to disk: 0.005144 seconds
vtk files anisoDelta_before written in 0.038241 seconds

MeshAdapt: ma::configure:  Based on requested sizefield, MeshAdapt requires at least 3 iterations.
           Setting the number of iteration to 4!

MeshAdapt: version 2.0 - dev !
planned Zoltan balance to target imbalance 1.100000 in 0.220910 seconds
Zoltan balanced to 1.100000 in 0.524734 seconds

MeshAdapt: iteration 0

MeshAdapt: coarsened 53 edges in 0.272583 seconds

MeshAdapt: refined 611639 edges in 3.049696 seconds

MeshAdapt: --iter 0 of shape correction loop: #bad elements 8683

MeshAdapt: --fixLargeAngles       in 0.430228 seconds: #bad elements 2235

MeshAdapt: --fixShortEdgeElements in 1.284403 seconds: #bad elements 818

MeshAdapt: --percent change in number of bad elements 0.905793

MeshAdapt: --iter 1 of shape correction loop: #bad elements 818

MeshAdapt: --fixLargeAngles       in 0.104712 seconds: #bad elements 288

MeshAdapt: --fixShortEdgeElements in 0.358577 seconds: #bad elements 143

MeshAdapt: --percent change in number of bad elements 0.825183

MeshAdapt: --iter 2 of shape correction loop: #bad elements 143

MeshAdapt: --fixLargeAngles       in 0.043554 seconds: #bad elements 82

MeshAdapt: --fixShortEdgeElements in 0.230717 seconds: #bad elements 58

MeshAdapt: --percent change in number of bad elements 0.594406

MeshAdapt: --iter 3 of shape correction loop: #bad elements 58

MeshAdapt: --fixLargeAngles       in 0.035118 seconds: #bad elements 56

MeshAdapt: --fixShortEdgeElements in 0.198542 seconds: #bad elements 53

MeshAdapt: --percent change in number of bad elements 0.086207

MeshAdapt: --iter 4 of shape correction loop: #bad elements 53

MeshAdapt: --fixLargeAngles       in 0.034678 seconds: #bad elements 52

MeshAdapt: --fixShortEdgeElements in 0.147253 seconds: #bad elements 49

MeshAdapt: --percent change in number of bad elements 0.075472

MeshAdapt: --iter 5 of shape correction loop: #bad elements 49

MeshAdapt: --fixLargeAngles       in 0.032862 seconds: #bad elements 52

MeshAdapt: --fixShortEdgeElements in 0.145992 seconds: #bad elements 49

MeshAdapt: --percent change in number of bad elements 0.000000

MeshAdapt: bad shapes down from 8683 to 49 in 31.149691 seconds

MeshAdapt: iteration 1

MeshAdapt: coarsened 10588 edges in 2.884405 seconds

MeshAdapt: refined 508001 edges in 4.174828 seconds

MeshAdapt: --iter 0 of shape correction loop: #bad elements 21415

MeshAdapt: --fixLargeAngles       in 0.773972 seconds: #bad elements 2934

MeshAdapt: --fixShortEdgeElements in 0.928731 seconds: #bad elements 556

MeshAdapt: --percent change in number of bad elements 0.974037

MeshAdapt: --iter 1 of shape correction loop: #bad elements 556

MeshAdapt: --fixLargeAngles       in 0.114474 seconds: #bad elements 150

MeshAdapt: --fixShortEdgeElements in 0.323925 seconds: #bad elements 71

MeshAdapt: --percent change in number of bad elements 0.872302

MeshAdapt: --iter 2 of shape correction loop: #bad elements 71

MeshAdapt: --fixLargeAngles       in 0.050748 seconds: #bad elements 40

MeshAdapt: --fixShortEdgeElements in 0.133650 seconds: #bad elements 32

MeshAdapt: --percent change in number of bad elements 0.549296

MeshAdapt: --iter 3 of shape correction loop: #bad elements 32

MeshAdapt: --fixLargeAngles       in 0.050780 seconds: #bad elements 28

MeshAdapt: --fixShortEdgeElements in 0.070426 seconds: #bad elements 28

MeshAdapt: --percent change in number of bad elements 0.125000

MeshAdapt: --iter 4 of shape correction loop: #bad elements 28

MeshAdapt: --fixLargeAngles       in 0.049759 seconds: #bad elements 32

MeshAdapt: --fixShortEdgeElements in 0.070114 seconds: #bad elements 28

MeshAdapt: --percent change in number of bad elements 0.000000

MeshAdapt: bad shapes down from 21415 to 28 in 38.095028 seconds

MeshAdapt: iteration 2

MeshAdapt: coarsened 145151 edges in 7.693994 seconds

MeshAdapt: refined 31824 edges in 1.503134 seconds

MeshAdapt: --iter 0 of shape correction loop: #bad elements 1294

MeshAdapt: --fixLargeAngles       in 0.195557 seconds: #bad elements 226

MeshAdapt: --fixShortEdgeElements in 0.222646 seconds: #bad elements 32

MeshAdapt: --percent change in number of bad elements 0.975270

MeshAdapt: --iter 1 of shape correction loop: #bad elements 32

MeshAdapt: --fixLargeAngles       in 0.045653 seconds: #bad elements 17

MeshAdapt: --fixShortEdgeElements in 0.112203 seconds: #bad elements 12

MeshAdapt: --percent change in number of bad elements 0.625000

MeshAdapt: --iter 2 of shape correction loop: #bad elements 12

MeshAdapt: --fixLargeAngles       in 0.045051 seconds: #bad elements 11

MeshAdapt: --fixShortEdgeElements in 0.066190 seconds: #bad elements 11

MeshAdapt: --percent change in number of bad elements 0.083333

MeshAdapt: --iter 3 of shape correction loop: #bad elements 11

MeshAdapt: --fixLargeAngles       in 0.047230 seconds: #bad elements 11

MeshAdapt: --fixShortEdgeElements in 0.057770 seconds: #bad elements 10

MeshAdapt: --percent change in number of bad elements 0.090909

MeshAdapt: --iter 4 of shape correction loop: #bad elements 10

MeshAdapt: --fixLargeAngles       in 0.044678 seconds: #bad elements 10

MeshAdapt: --fixShortEdgeElements in 0.057195 seconds: #bad elements 10

MeshAdapt: --percent change in number of bad elements 0.000000

MeshAdapt: bad shapes down from 1294 to 10 in 31.419985 seconds

MeshAdapt: iteration 3

MeshAdapt: coarsened 29195 edges in 3.057790 seconds

MeshAdapt: refined 4222 edges in 0.570225 seconds

MeshAdapt: --iter 0 of shape correction loop: #bad elements 158

MeshAdapt: --fixLargeAngles       in 0.100734 seconds: #bad elements 56

MeshAdapt: --fixShortEdgeElements in 0.192765 seconds: #bad elements 16

MeshAdapt: --percent change in number of bad elements 0.898734

MeshAdapt: --iter 1 of shape correction loop: #bad elements 16

MeshAdapt: --fixLargeAngles       in 0.044524 seconds: #bad elements 13

MeshAdapt: --fixShortEdgeElements in 0.056625 seconds: #bad elements 11

MeshAdapt: --percent change in number of bad elements 0.312500

MeshAdapt: --iter 2 of shape correction loop: #bad elements 11

MeshAdapt: --fixLargeAngles       in 0.046506 seconds: #bad elements 10

MeshAdapt: --fixShortEdgeElements in 0.056879 seconds: #bad elements 10

MeshAdapt: --percent change in number of bad elements 0.090909

MeshAdapt: --iter 3 of shape correction loop: #bad elements 10

MeshAdapt: --fixLargeAngles       in 0.044946 seconds: #bad elements 10

MeshAdapt: --fixShortEdgeElements in 0.055969 seconds: #bad elements 10

MeshAdapt: --percent change in number of bad elements 0.000000

MeshAdapt: bad shapes down from 158 to 10 in 23.725210 seconds

MeshAdapt: iteration 4

MeshAdapt: coarsened 4361 edges in 1.933079 seconds

MeshAdapt: refined 667 edges in 0.528031 seconds

MeshAdapt: --iter 0 of shape correction loop: #bad elements 55

MeshAdapt: --fixLargeAngles       in 0.099377 seconds: #bad elements 29

MeshAdapt: --fixShortEdgeElements in 0.114539 seconds: #bad elements 9

MeshAdapt: --percent change in number of bad elements 0.836364

MeshAdapt: --iter 1 of shape correction loop: #bad elements 9

MeshAdapt: --fixLargeAngles       in 0.043805 seconds: #bad elements 9

MeshAdapt: --fixShortEdgeElements in 0.060760 seconds: #bad elements 9

MeshAdapt: --percent change in number of bad elements 0.000000

MeshAdapt: bad shapes down from 55 to 9 in 11.790645 seconds

MeshAdapt: iteration 5

MeshAdapt: coarsened 665 edges in 1.713284 seconds

MeshAdapt: refined 164 edges in 0.503225 seconds

MeshAdapt: --iter 0 of shape correction loop: #bad elements 35

MeshAdapt: --fixLargeAngles       in 0.094527 seconds: #bad elements 28

MeshAdapt: --fixShortEdgeElements in 0.109800 seconds: #bad elements 9

MeshAdapt: --percent change in number of bad elements 0.742857

MeshAdapt: --iter 1 of shape correction loop: #bad elements 9

MeshAdapt: --fixLargeAngles       in 0.045398 seconds: #bad elements 9

MeshAdapt: --fixShortEdgeElements in 0.060482 seconds: #bad elements 9

MeshAdapt: --percent change in number of bad elements 0.000000

MeshAdapt: bad shapes down from 35 to 9 in 11.793450 seconds

MeshAdapt: iteration 6

MeshAdapt: coarsened 162 edges in 1.659875 seconds

MeshAdapt: refined 74 edges in 0.500956 seconds

MeshAdapt: --iter 0 of shape correction loop: #bad elements 30

MeshAdapt: --fixLargeAngles       in 0.093688 seconds: #bad elements 28

MeshAdapt: --fixShortEdgeElements in 0.109019 seconds: #bad elements 9

MeshAdapt: --percent change in number of bad elements 0.700000

MeshAdapt: --iter 1 of shape correction loop: #bad elements 9

MeshAdapt: --fixLargeAngles       in 0.043221 seconds: #bad elements 9

MeshAdapt: --fixShortEdgeElements in 0.059726 seconds: #bad elements 9

MeshAdapt: --percent change in number of bad elements 0.000000

MeshAdapt: bad shapes down from 30 to 9 in 11.783985 seconds

MeshAdapt: iteration 7

MeshAdapt: coarsened 62 edges in 1.573582 seconds

MeshAdapt: refined 35 edges in 0.505162 seconds

MeshAdapt: --iter 0 of shape correction loop: #bad elements 29

MeshAdapt: --fixLargeAngles       in 0.096616 seconds: #bad elements 28

MeshAdapt: --fixShortEdgeElements in 0.110339 seconds: #bad elements 9

MeshAdapt: --percent change in number of bad elements 0.689655

MeshAdapt: --iter 1 of shape correction loop: #bad elements 9

MeshAdapt: --fixLargeAngles       in 0.043220 seconds: #bad elements 9

MeshAdapt: --fixShortEdgeElements in 0.059952 seconds: #bad elements 9

MeshAdapt: --percent change in number of bad elements 0.000000

MeshAdapt: bad shapes down from 29 to 9 in 11.728326 seconds

MeshAdapt: iteration 8

MeshAdapt: coarsened 26 edges in 1.632900 seconds

MeshAdapt: refined 34 edges in 0.506708 seconds

MeshAdapt: --iter 0 of shape correction loop: #bad elements 28

MeshAdapt: --fixLargeAngles       in 0.044710 seconds: #bad elements 28

MeshAdapt: --fixShortEdgeElements in 0.110796 seconds: #bad elements 9

MeshAdapt: --percent change in number of bad elements 0.678571

MeshAdapt: --iter 1 of shape correction loop: #bad elements 9

MeshAdapt: --fixLargeAngles       in 0.043211 seconds: #bad elements 9

MeshAdapt: --fixShortEdgeElements in 0.059862 seconds: #bad elements 9

MeshAdapt: --percent change in number of bad elements 0.000000

MeshAdapt: bad shapes down from 28 to 9 in 11.647666 seconds

MeshAdapt: iteration 9

MeshAdapt: coarsened 19 edges in 1.560987 seconds

MeshAdapt: refined 24 edges in 0.507650 seconds

MeshAdapt: --iter 0 of shape correction loop: #bad elements 28

MeshAdapt: --fixLargeAngles       in 0.045709 seconds: #bad elements 28

MeshAdapt: --fixShortEdgeElements in 0.110187 seconds: #bad elements 9

MeshAdapt: --percent change in number of bad elements 0.678571

MeshAdapt: --iter 1 of shape correction loop: #bad elements 9

MeshAdapt: --fixLargeAngles       in 0.044045 seconds: #bad elements 9

MeshAdapt: --fixShortEdgeElements in 0.059999 seconds: #bad elements 9

MeshAdapt: --percent change in number of bad elements 0.000000

MeshAdapt: bad shapes down from 28 to 9 in 11.919131 seconds

MeshAdapt: Maximum (metric) edge length in the mesh is 1.709849

MeshAdapt: 0th additional refine-snap call

MeshAdapt: refined 19 edges in 0.540297 seconds

MeshAdapt: Maximum (metric) edge length in the mesh is 1.499999

MeshAdapt: worst element quality is 1.697484e-02

MeshAdapt: worst element quality is 1.697484e-02
planned Zoltan balance to target imbalance 1.100000 in 2.036767 seconds
Zoltan balanced to 1.100000 in 7.346896 seconds

MeshAdapt: element imbalance 9% of average

MeshAdapt: mesh adapted in 308.030375 seconds
number of tet 5934317 hex 0 prism 0 pyramid 0
mesh entity counts: v 1067369 e 7026323 f 11893272 r 5934317
  - verifying tags: mach_ver, original_metric_ver, target_metric_ver
mesh verified in 1.533155 seconds

MeshAdapt: ma::configure:  Based on requested sizefield, MeshAdapt requires at least 1 iterations.
           Setting the number of iteration to 2!
writeVtuFile into buffers: 0.436119 seconds
writeVtuFile buffers to disk: 0.027537 seconds
vtk files anisoDelta_after written in 0.478725 seconds
```
</details>

cwsmith:
The timing and counts look good.  Here are the final counts from my old run on piglet:

```
MeshAdapt: mesh adapted in 450.370630 seconds
number of tet 5934317 hex 0 prism 0 pyramid 0
mesh entity counts: v 1067369 e 7026323 f 11893272 r 5934317
  - verifying tags: mach_ver, original_metric_ver, target_metric_ver
mesh verified in 1.516892 seconds
```

bobpaw:
Ok, great! Anything else I need to do before PR is approved?

jacobmerson:
All set from my end.

pcu/PCU.cc
```diff
@@ -52,6 +52,23 @@
 #include <time.h>
 namespace pcu {
 
+int PCU_Init(int *argc, char ***argv) {
+#ifndef SCOREC_NO_MPI
+  return MPI_Init(argc, argv);
```
cwsmith, line 57: We should check if MPI was already initialized here (https://www.mpich.org/static/docs/latest/www3/MPI_Initialized.html).  Initializing multiple times is undefined behavior or 'wrong' (https://docs.open-mpi.org/en/v5.0.x/man-openmpi/man3/MPI_Init.3.html#mpi-init, https://www.mpich.org/static/docs/v3.3/www3/MPI_Init.html).

pcu/PCU.cc
```diff
@@ -52,6 +52,23 @@
...
+#else
+  (void) argc, (void) argv;
+  return 0;
+#endif
+}
+
+int PCU_Finalize() {
+#ifndef SCOREC_NO_MPI
+  return MPI_Finalize();
```
cwsmith, line 66: As above, I think we should check if MPI was already finalized before calling finalize ourselves.  https://www.mpich.org/static/docs/latest/www3/MPI_Finalized.html

cwsmith (changes requested):
Looks good.  Thank you. I left a couple comments below.

pcu/PCU.cc
bobpaw: Should pcu::PCU_Init throw an exception in this case?

bobpaw:
Do you think I should rename pcu::PCU_Init to pcu::Init? same suggestion for finalize. The C api I would keep the same. I think it's still namespaced and sorter is nicer.

pcu/PCU.cc
cwsmith: Nope, we just don't call initialize.

cwsmith:
> Do you think I should rename pcu::PCU_Init to pcu::Init? same suggestion for finalize. The C api I would keep the same. I think it's still namespaced and sorter is nicer.

Either way is fine with me.

Another change request: we need to increase the pumi version number to 4.0.0 in CMakeLists.txt as we have changed the api (https://semver.org/).

pcu/PCU.cc
cwsmith: That works for me.  Thanks for checking.

bobpaw:
> > Do you think I should rename pcu::PCU_Init to pcu::Init? same suggestion for finalize. The C api I would keep the same. I think it's still namespaced and sorter is nicer.
> 
> Either way is fine with me.
> 
> Another change request: we need to increase the pumi version number to 4.0.0 in CMakeLists.txt (
> 
> https://github.com/SCOREC/core/blob/17ad033da30c8079a69e8142486ce01892c562b8/CMakeLists.txt#L9
> ) as we have changed the api (https://semver.org/).

Is there anything we want to bundle with this major release?

cwsmith:
> Is there anything we want to bundle with this major release?

Not that I can think of.

cwsmith:
@bobpaw The CI tests caught what I'm guessing is a missing header?

```
 /space/smithc11/githubSelfHostedRunner/actions-runner/_work/core/core/core_425/test/sim_countBL.cc: In function 'int main(int, char**)':
/space/smithc11/githubSelfHostedRunner/actions-runner/_work/core/core/core_425/test/sim_countBL.cc:25:3: error: 'pcu' has not been declared
   25 |   pcu::Init(&argc, &argv);
      |   ^~~
/space/smithc11/githubSelfHostedRunner/actions-runner/_work/core/core/core_425/test/sim_countBL.cc:110:3: error: 'pcu' has not been declared
  110 |   pcu::Finalize();
      |   ^~~
```

bobpaw:
I think I had to edit that file in the deltaWing example. That file is also using MPI allreduce, peers, etc. so I will need to change a few other lines.

cwsmith (approved):

cwsmith (changes requested):
Whooops.  I misread the CI test results.  We need to fix the cgns tests.

bobpaw:
@cwsmith would you do the honors?

cwsmith (approved):

PR #427, opened 2024-05-08
fix: Replace sprintf with snprintf

bobpaw:
## Replace sprintf with snprintf

Fixed potential buffer overflows given long filenames.


cwsmith (approved):
LGTM.  Thank you.

Issue #434, opened 2024-06-14
ctest failures with simmodsuite 2024.1-240606dev


cwsmith:
There are test failures when building against the latest Simmetrix SimModSuite development version: 2024.1-240606dev

## SCOREC RHEL9 environment

```
module use /opt/scorec/spack/rhel9/v0201_4/lmod/linux-rhel9-x86_64/Core/
module load gcc/12.3.0-iil3lno mpich/4.1.1-xpoyz4t cmake
module load zoltan
module load simmetrix-simmodsuite/2024.1-240606dev-pi7eze2
```

## cmake command and output

```
$ cmake -S /space/cwsmith/pumiTestRhel9Sim/core -B buildPumiOptonSimonOmegaoff -DBUILD_SHARED_LIBS=off -DCMAKE_C_COMPILER=mpicc -DCMAKE_CXX_COMPILER=mpicxx -DCMAKE_Fortran_COMPILER=gfortran -DSCOREC_CXX_OPTIMIZE=on -DSCOREC_EXTRA_CXX_FLAGS= -DVALGRIND= -DVALGRIND_ARGS= -DMDS_ID_TYPE=int -DPCU_COMPRESS=ON -DENABLE_ZOLTAN=ON -DPUMI_FORTRAN_INTERFACE=off -DENABLE_OMEGA_H=off -DIS_TESTING=ON -DENABLE_SIMMETRIX=on -DSKIP_SIMMETRIX_VERSION_CHECK=on -DENABLE_FIELDSIM=OFF -DSIM_M
PI=mpich4.1.1 -DSIM_PARASOLID=on -DSIM_ACIS=on -DMESHES=/space/cwsmith/pumiTestRhel9Sim/core/pumi-meshes/ -DCMAKE_INSTALL_PREFIX=buildPumiOptonSimonOmegaoff/install                                                                                          
                                                                                                                               
-- The CXX compiler identification is GNU 12.3.0                                                                                                                                                                                                                                                                                                                                             
-- The C compiler identification is GNU 12.3.0                                                                                                                                                
-- Detecting CXX compiler ABI info                                                                                                                                                                                                                                                                                                                                                           
-- Detecting CXX compiler ABI info - done                                                                                                                                                                                                                                                                                                                                                    
[...]

PR #435, opened 2024-06-21
Add CTest DEPENDS property

bobpaw:
- CMake test property DEPENDS has been added to the appropriate tests to facilitate parallel testing.
- Some tests require other tests to run first to operate on their output (split -> collapse, etc). Previously, tests would fail once and then succeed on a second ctest run. Now in parallel those tests wait for what they need, allowing tests to pass the first time.

### DIFF
--- a/test/testing.cmake
+++ b/test/testing.cmake
@@ -264,6 +264,8 @@ mpi_test(inviscid_ghost 4
   "${MDIR}/inviscid_egg.dmg"
   "${MDIR}/4/"
   "${MDIR}/vis")
+set_tests_properties(inviscid_ghost PROPERTIES DEPENDS inviscid_ugrid)
+
 set(MDIR ${MESHES}/pipe)
 if(ENABLE_SIMMETRIX)
   mpi_test(convert 1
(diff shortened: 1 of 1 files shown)


bobpaw:
Some Simmetrix tests are failing but they were also failing before these changes.

cwsmith:
@bobpaw Thank you for the changes.  The failing tests are the same as those listed here: https://github.com/SCOREC/core/issues/434

Does the CMake minimum version: https://github.com/SCOREC/core/blob/d68636eb29e360774ecd73c820ca933b190321f5/CMakeLists.txt#L7 need to be increased to support the DEPENDS property?

bobpaw:
> @bobpaw Thank you for the changes. The failing tests are the same as those listed here: #434
> 
> Does the CMake minimum version:
> 
> https://github.com/SCOREC/core/blob/d68636eb29e360774ecd73c820ca933b190321f5/CMakeLists.txt#L7
> need to be increased to support the DEPENDS property?

No, based on this documentation it has been around since CMake 3.0:
https://cmake.org/cmake/help/v3.0/prop_test/DEPENDS.html

cwsmith (approved):

PR #436, opened 2024-06-21
Use snprintf in tests

bobpaw:
- I missed some sprintf calls before. I don't see any others (find + grep).
- These deprecation errors come up with LLVM so others might not notice.

### DIFF
--- a/test/viz.cc
+++ b/test/viz.cc
@@ -65,7 +65,7 @@ int main(int argc, char** argv)
 
     }
     v.markPart(m,part_num);
-    snprintf(output,256,"Testing MIS %d",i);
+    snprintf(output,128,"Testing MIS %d",i);
     v.breakpoint(std::string(output));
   }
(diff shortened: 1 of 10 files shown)


test/viz.cc
```diff
@@ -65,7 +65,7 @@ int main(int argc, char** argv)
 
     }
     v.markPart(m,part_num);
-    sprintf(output,"Testing MIS %d",i);
+    snprintf(output,256,"Testing MIS %d",i);
```
cwsmith, line 68: Should this be 128?

cwsmith (changes requested):
Thank you for the changes.  I may have found one size mismatch.

test/viz.cc
bobpaw: Yep. Fixed it.

cwsmith (approved):

Issue #440, opened 2024-07-26
getElementNodeData assumes scalar field


jacobmerson:
The function here:
https://github.com/SCOREC/core/blob/8959c599cc05e21d0fb470f941d9f892da62aa02/apf/apfElement.cc#L122-L127

Assumes that the field stored at the nodes is a scalar.

jacobmerson:
Closed based on #443

PR #442, opened 2024-07-28
Update apfElement.cc to fix bug described in #440

jacobmerson:
This pull request fixes the bug in #440 so that getElementNodeData works with fields that have any number of components.


cwsmith:
@jacobmerson Thank you.  Would you please rebase this on develop?

cwsmith:
replaced by #443

Issue #444, opened 2024-08-09
Expensive gmi_is_point_in_region called during ma::adapt


bobpaw:
I'm reporting an issue in ma::adapt where an unnecessarily expensive call (depending on geometric complexity) is made to gmi_is_point_in_region.

Call stack (from gdb):
```
#0  gmi_is_point_in_region (m=0x8df910, e=0xb, point=0x7ffffffed030)
    at gmi/gmi.c:143
#1  0x00007ffff7cc366d in apf::Mesh::isParamPointInsideModel (this=this@entry=0x7f9d80,
    g=g@entry=0x2a, param=..., x=...)
    at apf/apfMesh.cc:271
#2  0x00007ffff7e77589 in ma::interpolateParametricCoordinatesOnRegularFace (p=..., b=...,
    a=..., t=0.5, g=0x2a, m=0x7f9d80)
    at ma/maSnap.cc:500
#3  ma::interpolateParametricCoordinatesOnFace (p=..., b=..., a=..., t=0.5, g=0x2a,
    m=0x7f9d80) at ma/maSnap.cc:570
#4  ma::interpolateParametricCoordinates (m=0x7f9d80, g=0x2a, t=0.5, a=..., b=..., p=...)
    at ma/maSnap.cc:586
#5  0x00007ffff7e7781b in ma::transferParametricBetween (p=..., t=0.5, v=0x7ffffffed0c0,
    g=0x2a, m=0x7f9d80) at ma/maSnap.cc:601
#6  ma::transferParametricOnEdgeSplit (m=m@entry=0x7f9d80, e=e@entry=0xc9ea2, t=t@entry=0.5,
    p=...) at ma/maSnap.cc:615
#7  0x00007ffff7e5fecb in ma::makeSplitVert (r=r@entry=0x6f6d550, edge=edge@entry=0xc9ea2)
    at ma/maRefine.cc:144
#8  0x00007ffff7e62d8a in ma::splitEdge (r=0x6f6d550, edge=0xc9ea2, v=0x7ffffffed240)
    at ma/maTemplates.cc:81
#9  0x00007ffff7e601af in ma::splitElement (r=r@entry=0x6f6d550, e=e@entry=0xc9ea2)
    at ma/maRefine.cc:244
#10 0x00007ffff7e60643 in ma::splitElements (r=r@entry=0x6f6d550)
    at ma/maRefine.cc:334
#11 0x00007ffff7e60dbd in ma::refine (a=a@entry=0x816270)
[...]

PR #445, opened 2024-08-09
Fix logic checking periodic model faces

joshia5:
Fix bug in logic checking periodic model faces when computing parametric coordinates

addressing issue https://github.com/SCOREC/core/issues/444


joshia5:
closing this pr because I accidentally base from master,  correct one is here https://github.com/SCOREC/core/pull/446

PR #446, opened 2024-08-09
Fix logic checking periodic model faces

joshia5:
Fix bug in logic checking periodic model faces when computing parametric coordinates

This PR is relating to issue https://github.com/SCOREC/core/issues/444

The PR is correcting the logic that checks if case is periodic and then only execute the lines https://github.com/SCOREC/core/blob/60c507f0c2499b31419369683c4ddfecacc68266/ma/maSnap.cc#L502-L511 that address periodic cases

### DIFF
--- a/ma/maSnap.cc
+++ b/ma/maSnap.cc
@@ -486,7 +486,7 @@ static void interpolateParametricCoordinatesOnRegularFace(
 
   /* check if the new point is inside the model.
    * otherwise re-run the above loop with last option
-   * in "interpolateParametricCoordinae" being 1.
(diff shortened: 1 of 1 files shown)


bobpaw:
@joshia5 @cwsmith, is this ready to be merged?

cwsmith:
@bobpaw I'm running tests again (which now includes CGNS tests).  If those pass I have no objections to merging this.

cwsmith:
@joshia5 Would you mind rebasing this on develop?  The CGNS tests are failing as this branch is missing some fixes.

PR #451, opened 2024-08-29
Set bob install dir to CMAKE_INSTALL_LIBDIR

bobpaw:
## Set bob install dir to CMAKE_INSTALL_LIBDIR

- Fixes issues on 64bit systems when LIBDIR=lib64. CMake would relink to $ORIGIN:$ORIGIN/../lib64 but all the libs would be in lib.

### DIFF
--- a/cmake/bob.cmake
+++ b/cmake/bob.cmake
@@ -173,7 +173,7 @@ endmacro(bob_end_subdir)
 
 function(bob_end_package)
   include(CMakePackageConfigHelpers)
-  set(INCLUDE_INSTALL_DIR include)
+  set(INCLUDE_INSTALL_DIR "${CMAKE_INSTALL_INCLUDEDIR}")
   set(LIB_INSTALL_DIR "${CMAKE_INSTALL_LIBDIR}")
   set(CONFIG_CONTENT "
 set(${PROJECT_NAME}_VERSION ${${PROJECT_NAME}_VERSION})
--- a/cmake/bob.cmake
+++ b/cmake/bob.cmake
@@ -155,11 +155,11 @@ endmacro(bob_public_dep)
 
 function(bob_export_target tgt_name)
   install(TARGETS ${tgt_name} EXPORT ${tgt_name}-target
-      RUNTIME DESTINATION bin
-      ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
-      LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR})
+      RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
+      ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
+      LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}")
   install(EXPORT ${tgt_name}-target NAMESPACE ${PROJECT_NAME}::
-          DESTINATION lib/cmake/${PROJECT_NAME})
+          DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}")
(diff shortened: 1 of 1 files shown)


cmake/bob.cmake
```diff
@@ -156,8 +156,8 @@ endmacro(bob_public_dep)
 function(bob_export_target tgt_name)
   install(TARGETS ${tgt_name} EXPORT ${tgt_name}-target
       RUNTIME DESTINATION bin
```
cwsmith, line 158: I think we should also set this to `CMAKE_INSTALL_BINDIR` to keep things consistent.

https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

cwsmith (changes requested):
Looks good.  Thank you.

What is the issue with relinking and `$ORIGIN`?  Is that for capstone?

I see a handful of locations in bob.cmake where `lib` is used explicitly.  I'm guessing those should also be changed.

bobpaw:
> Looks good. Thank you.
> 
> What is the issue with relinking and `$ORIGIN`? Is that for capstone?

Yes. Capstone builds SCOREC/core and provides the shared libraries with capstone. That capstone library can be linked again to SCOREC (which will be especially necessary if the plugin is provided by default), but the issue I was having was that ld would find the shared libraries (ma.so, mds.so, gmi.so) in the capstone folder instead of the ones I had installed and built (with `cmake --install`). Using the build tree was easy, but trying to use the install tree was having this `lib64` issue due to relinking.

> 
> I see a handful of locations in bob.cmake where `lib` is used explicitly. I'm guessing those should also be changed.

I believe those are only CMake targets. I don't know the convention, but the Capstone package includes those in `lib`.

cwsmith:
Thank you.  I think I get the gist of the capstone problem.

It looks like the Professional CMake book (19th edition) suggests having a consistent directory for the libraries, <packageName>Config.cmake, and target files.

> The install destination of that file (<packageName>Config.cmake)  should use the variables defined by the GNUInstallDirs, which simplifies customization by Linux distributions and other packaging systems. [pg 541]

I didn't find a direct quote for the target files, but all the examples I could find (see pg 524) use the GNUInstallDirs variables.

I can help with these modifications and testing if needed.

bobpaw:
Ok, that makes sense. I will update the remaining `lib` to be `CMAKE_INSTALL_LIBDIR`. I will also update `bin` per the same guideline.

cmake/bob.cmake
```diff
@@ -174,7 +174,7 @@ endmacro(bob_end_subdir)
 function(bob_end_package)
   include(CMakePackageConfigHelpers)
   set(INCLUDE_INSTALL_DIR include)
```
bobpaw, line 176: I do wonder if this `include` should be replaced with CMAKE_INSTALL_INCLUDEDIR or CMAKE_INSTALL_OLDINCLUDEDIR? Not sure if that's overkill since honestly that path likely won't change.

https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

cwsmith: I think using `CMAKE_INSTALL_INCLUDEDIR` here is the clean/consistent approach.  Thank you.

cwsmith:
Thank you. I'm going to add a CI test to check that downstream packages can use our cmake package successfully.

cwsmith:
@bobpaw Can you rebase on develop?  It now has a CI test for a user build using the installed cmake config file?

https://github.com/SCOREC/core/blob/88a39f6c4db82e82595c47cfc7fdfb56cbf3fec0/.github/workflows/cmake.yml#L48-L56

cwsmith:
Thank you.  The new test ran without issues.

https://github.com/SCOREC/core/actions/runs/10755354998/job/29826921049?pr=451#step:7:1

PR #453, opened 2024-09-13
Fix unused variable warning introduced by #446

bobpaw:
## Fix unused variable warning introduced by #446

This fixes an unused warning when CMake is given `-DENABLE_CAPSTONE` by casting to void.

However, I wonder if we can remove the `#ifndef` altogether?


bobpaw:
@onkarsahni Do you have any insights in terms of Capstone and periodic meshes? Is that something they might support in the future? If we remove the `#ifndef HAVE_CAPSTONE` the expensive parametric call still won't be made unless it's a 2d periodic mesh.

PR #459, opened 2024-10-07
Clarify that orthogonal basis frames are columns

bobpaw:
- Although apf::Matrix stores rows of vectors, MeshAdapt expects basis frames on columns (as indicated by ma/maSize.cc:97). This was not made clear anywhere else. I wasted time not realizing it, so I added it explicitly in header documentation instead of buried in source code.


joshia5 (approved):
Thank you, this LGTM.

PR #466, opened 2024-11-08
test globus

Angelyr:
- Moves all of the files that globus uses unto Github.
- Moves globus code to separate repository so that it can be reused.


cwsmith (approved):
LGTM.  Thank you.

PR #472, opened 2025-02-06
add ctest PROCESSORS property to aid scheduling

bobpaw:
## add ctest PROCESSORS property to aid scheduling

This helps speed up testing when using `ctest -j8` or the like, because the CTest scheduler can run groups of tests based on the processors they use. Without this info, it may schedule 8 tests at once, some of which actually use multiple processors.

- test/testing.cmake: add PROCESSORS property to tests created with mpi_test to help the parallel scheduler.

On trouble, I got the following results:

without PROCESSORS: `Total Test time (real) =  94.14 sec`
with PROCESSORS: `Total Test time (real) =  60.01 sec`


cwsmith (approved):
Nice.  Thank you.

Issue #480, opened 2025-04-24
Replace deprecated `std::iterator`


Sichao25:
`std::iterator`  is set to be deprecated in C++17. See more explanations [here](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0174r1.html#2.1). Its usage should be replaced to ensure compatibility with newer C++ standards.

bobpaw:
If we're making changes to support later C++ standards, I would like to see APF_ITERATE deprecated. If there are internal uses they should probably be range-based for loops.

jacobmerson:
I played around with the idea of proper iterator support for PUMI meshes a long time ago. There is one subtle point that makes that tricky. The range-based for loops rely on `begin` and `end`. However, for a mesh, we may want to loop over all dimensions, or just a single dimension. I'm sure there is some interface that makes sense and can support that I just am not 100% sure what it looks like.

jacobmerson:
@bobpaw is there a strong reason to replace these? It seems like it will be more effort than it's worth to remove it.

bobpaw:
Nope, just a wish-list item.

cwsmith:
closed with https://github.com/SCOREC/core/pull/482

PR #482, opened 2025-05-01
Remove deprecated `std::iterator`

Sichao25:
Solved the warning related to #480 . `std::iterator` is set to be deprecated in C++17. We are expected to define the types exposed by our iterators.


cwsmith (approved):
Looks good.  Thank you.

PR #483, opened 2025-05-06
Enable cxx standard 20 test cases in CI test matrix

Sichao25:
Explicitly configure the CI workflow to test with both C++11 (the default standard when building without CGNS) and C++20.


cwsmith:
Thank you @Sichao25.

It looks like this is hitting the known issue with deprecated std::iterator (https://github.com/SCOREC/core/actions/runs/14848813109/job/41688398268?pr=483#step:5:744) as discussed in #480 .  If that is the only issue, then I think we are good to go.

Were you planning on other changes?

cwsmith:
@Sichao25 Before merging this I think we should push a fix for the deprecated std::iterator so the llvm build doesn't fail with C++20.

PR #486, opened 2025-05-29
Add capAdapt test utility and framework_meshing C++14 requirement

bobpaw:
## Add capAdapt test utility and framework_meshing C++14 requirement

- Add `capAdapt` test utility to load Capstone mesh, convert to MDS, partition, adapt, join, convert back to and write CRE.
- Update CMake linking for `framework_meshing`: the headers use C++14 so it needs the `INTERFACE` compile feature.
- Link `capAdapt` against `framework_meshing` to call smoothing directly and avoid converting to frame/scale fields between sizing.
    - Thinking about moving the code to import from bulk sizing file (possibly with smoothing) directly into apf_cap.


cwsmith (approved):
LGTM.

PR #490, opened 2025-06-10
Fix PCU Pack argument constness, remove pointer overloads, and add documentation

bobpaw:
## Fix PCU Pack argument constness, remove pointer overloads, and add documentation

- add const to PCU::Pack data argument. The underlying call has a `const void*` argument, so without const on the 2-argument overload it is not possible to pass rvalues (temporary objects).
- remove PCU::Pack and PCU::Unpack overloads with `T*& data`. They were only used to send `MeshEntity*`s which are typedefs for integral types. The function body was the same for both overloads (`T&` and `T*&`), so removing the more specific one (`T*&`) results in no change in behavior.
- add and update documentation.
- move PCU Doxygen `\page` to pcu/pcu.dox (which follows the style of apf).
- add other public PCU headers to Doxygen `INPUT` for the future.


pcu/PCU.h
```diff
@@ -19,23 +32,20 @@ class PCU {
   PCU(PCU &&) noexcept;
   PCU &operator=(PCU const &) = delete;
   PCU &operator=(PCU &&) noexcept;
-  /** @brief Returns the rank of the current process.
-   *  @return The rank of the current process.
+  /** \brief Returns the rank of the current process.
```
cwsmith, line 35: Is the change from `@` to `\` required?

cwsmith (commented):
Looks good.  Thank you.

PR #495, opened 2025-06-13
Update python wrappers

Sichao25:
The Python wrappers have been out of date for some time and were only compatible with older versions of PUMI. This PR addresses several related issues:
- Updates deprecated APIs in the interface files.
- Comment out a few lines in the preprocessor to avoid a parsing problem in SWIG (other workarounds may also be possible).
- Updates the Python test scripts accordingly.
- Adds a `.smd` file as an alternative to `.x_t` to adapt to different versions of Simmetrix.
- Adds a new CI workflow that builds and runs the Python test script without requiring Simmetrix.

### DIFF
--- a/pcu/PCU.h
+++ b/pcu/PCU.h
@@ -110,6 +110,11 @@ public:
   int Received(size_t *size) noexcept;
   void *Extract(size_t size) noexcept;
 
+  /*
+   * Debug print function with printf-style formatting
+   * This function is excluded from SWIG parsing due to complex macro usage
+   * with variadic argument. Also, such functionalities is not needed for the Python API.
+   */
   #ifndef SWIG
   void DebugPrint(const char* format, ...) noexcept PCU_FORMAT_ATTRIBUTE(2, 3)
   void DebugPrint(const char* format, va_list args) noexcept;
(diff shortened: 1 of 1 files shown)


cwsmith:
Thank you @Sichao25 .  Does the following SO post help?
https://stackoverflow.com/questions/11403520/preprocessor-macro-in-swig

Sichao25:
Unfortunately, that SO doesn't work on my side. However, I was able to get things working by ignoring the SWIG unrecognized code during compilation and implementing it in another way in the SWIG interface file. (Thanks to Aiden for the idea and resources!)

I've added the required template initialization to the PR. I could also implement a similar `DebugPrint` function using this method. For example:
```
%extend pcu::PCU{
  void DebugPrint(FILE *f, const char *format, ...) {
    va_list args;
    va_start(args, format);
    vfprintf(f, format, args);
    fflush(f);
    va_end(args);
  }
}
```
But I'm not sure if it's worth the effort, since I feel like it would be relatively straightforward to achieve the same functionality in Python. 

Do you think it makes sense to skip the `DebugPrint` feature in the wrapped Python API, especially since it wasn't included in the older version either?

jacobmerson:
My opinion is skip `DebugPrint` whomever is using the python interface is just going to directly call `print` anyways.

jacobmerson:
The changes look good to me and the example is pretty clean.

Sichao25:
@cwsmith Does the comment make sense to you?

cwsmith (approved):
Looks good to me.  Thank you.

PR #498, opened 2025-06-19
Update doxygen configuration files

bobpaw:
## Update doxygen configuration files

- Configure public documentation `INPUT` programmatically from `HEADERS` lists in CMake `scorec_export_library`.
- Configure private documentation `INPUT` programmatically from subdirectories.
- Strip the CMake source path from all public documentation file names.
    - not sure why the default "shortest unique path" functionality wasn't working
    - Leave full filenames in for internal documentation because it may be important or include build details.
- Add `__cplusplus` to Doxygen `PREDEFINED` macros to ensure all APIs are extracted.
- Change Doxyfile_internal.in output directory so both documentations can be generated side by side.


cwsmith (approved):
This is a nice improvement.  Thank you.

PR #499, opened 2025-06-23
Fix balancing empty parts with metis

bobpaw:
## Fix balancing empty parts with metis

- added test mbalanceEmpty.cc to confirm that balancing with empty parts works.
- metis/apfMETISbalancer.cc: fixed helper function preconditions.
- (MetisBalancer::balance): allow balancing with one part (no work to do anyway).

### DIFF
--- a/test/mbalanceEmpty.cc
+++ b/test/mbalanceEmpty.cc
@@ -41,6 +41,7 @@ int main(int argc, char* argv[]) {
     } catch (...) {
       std::throw_with_nested(std::runtime_error("balancing failed"));
     }
+    m->verify();
     m->writeNative(argv[4]);
     // destroy mds
     m->destroyNative();
--- a/test/mbalanceEmpty.cc
+++ b/test/mbalanceEmpty.cc
@@ -1,0 +1,54 @@
+#include <iostream>
+#include <exception>
+#include <memory>
+#include <string>
+#include <apf.h>
+#include <apfMDS.h>
+#include <apfMesh2.h>
+#include <apfPartition.h>
+#include <gmi_mesh.h>
+#include <apfMETIS.h>
+#include <lionPrint.h>
+#include <pcu_util.h>
+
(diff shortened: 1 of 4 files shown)


test/mbalanceEmpty.cc
```diff
@@ -0,0 +1,55 @@
...
+#include <lionPrint.h>
+#include <pcu_util.h>
+
+int main(int argc, char* argv[]) {
+  pcu::Init(&argc, &argv);
+  try {
+    pcu::PCU PCU;
+    if (argc != 5) {
+      if (PCU.Self() == 0)
+        std::cerr << "USAGE: <model.dmg> <mesh.smb> <inParts> <out.smd>"
```
cwsmith, line 20:
```suggestion
        std::cerr << "USAGE: <model.dmg> <mesh.smb> <inParts> <outMesh.smb>"
```

test/mbalanceEmpty.cc
```diff
@@ -0,0 +1,55 @@
...
+    int inParts = std::stoi(argv[3]);
+    int group = PCU.Self() / inParts;
+    auto loadPCU = PCU.Split(group, 0);
+    gmi_model* model = gmi_load(argv[1]);
+    apf::Mesh2* m = nullptr;
+    if (group == 0) {
+      m = apf::loadMdsMesh(model, argv[2], loadPCU.get());
+      m->switchPCU(&PCU);
+    }
+    m = apf::expandMdsMesh(m, model, inParts, &PCU);
```
cwsmith, line 36: In the added test case, this is taking the single part pipe mesh and 'expanding' it to a four part mesh which has three empty parts?

bobpaw: Yes that is correct. I want to test METIS balancing with 1 or more empty parts.

I will add an assert/error message into the test case ensuring that `inParts < PCU.Peers()`.

PR #517, opened 2025-09-11
Rename CMake options and C macros to use PUMI_ prefix

### LINKED ISSUE #337: prefix cmake flags that enable optional features with PUMI
see https://github.com/SCOREC/core/pull/330#pullrequestreview-627429873

### LINKED ISSUE #477: Standardize macro names as part of next major release
I am proposing that we standardize macro names as part of the next major release.

Specifically: should everything be `PUMI_` or `SCOREC_` (in this case, the actual pumi library could still have `PUMI_`).

Personally I do not have a strong preference but I would like more consistency. Once a decision is made, we should update STYLE.md.

bobpaw:
## Rename CMake options and C macros to use PUMI_ prefix

I prefixed lots of options with PUMI_ and renamed variables/options with the SCOREC_ prefix which are not created by bob.cmake. In general I set the default value for options like `PUMI_ENABLE_SIMMETRIX` to `${ENABLE_SIMMETRIX}` which allows us to remain somewhat backward compatible and also inherit the selection from parent projects by default. 

To find things that needed to be changed, I used the following searches:

```
git grep '\WENABLE' :^Doxyfile.in :^Doxyfile_internal.in
git grep '\WHAS' :^Doxyfile.in :^Doxyfile_internal.in
git grep '\WHAVE' :^Doxyfile.in :^Doxyfile_internal.in
git grep '\WUSE' :^Doxyfile.in :^Doxyfile_internal.in
```

I removed the private compile definition `-DHAVE_SIMADVMESHING` from phasta/CMakeLists.txt and some test exes and replaced it with the `PUMI_HAS_SIMADVMESHING` macro in apf_sim/apf_simConfig.h. There are some options which are left, specifically Simmetrix things like `SIM_PARASOLID`, `SIM_ACIS`, etc that I wasn't sure how we wanted to deal with yet.

~~An exception that I did not remove is ENABLE_VIZ in test/CMakeLists.txt:~~ Removed in #531.

I did not yet change the project name to SCOREC because I wanted to make sure all of these changes work.

Leaving in draft mode for now because there are a few things (Phasta ~~HAVE_CLOCK_/USE_PCU_TIME~~, LION_COMPRESS, etc) left that I did not get to or am not sure if need to be updated.

Closes #337. Closes #477.


bobpaw:
I found what seems like a reasonable solution on the CMake discourse forum:

https://discourse.cmake.org/t/exporting-renamed-multiple-named-configs/2586

For us, the recommendation is to distribute a second `SCORECConfig.cmake` which exports all our targets pointing from `SCOREC::<target>` to `PUMI::<target>` depending on the requested version of PUMI.

jacobmerson:
This seems like a reasonable approach. Just came to mind, but do we want to have a specific target version or timespan where support of old naming will be dropped? Even if that doesn't go into the messages, it would be good for us to have an agreed upon/documented idea of when to do this.

bobpaw:
The minimum semantic versioning requirement for deprecation is to release a backwards compatible minor version with the deprecation, then a major version with the deprecated elements removed. It's also common to put these changes into a major 5.0 release and later release 6.0 with the old names dropped. I do not know what Cameron would prefer there or with respect to a timeline.

cwsmith:
Thank you @bobpaw .  

IIUC, as is (012d184), this PR is backwards compatible.  Is that correct?
If so, do we want to include the project name change (`SCOREC` -> `PUMI`) in this PR, create the backward compatible `SCORECConfig.cmake`, and bump the minor version number to 4.3.0?  As you described above, after some time we can then drop the old project name and create the 5.0 release.  

Note, we already have a 'downstream' proxy app we can test with in the repo here: https://github.com/SCOREC/core/tree/master/doc

bobpaw:
> IIUC, as is ([012d184](https://github.com/SCOREC/core/commit/012d184b841ce173ca4ba467544e54019c866227)), this PR is backwards compatible. Is that correct?

That is the intention, and I believe it is correct. Will need to do a final review to confirm.

bobpaw:
> If so, do we want to include the project name change (`SCOREC` -> `PUMI`) in this PR, create the backward compatible `SCORECConfig.cmake`, and bump the minor version number to 4.3.0? As you described above, after some time we can then drop the old project name and create the 5.0 release.

Sounds good. I will get on that.

PR #520, opened 2025-11-13
Fix Failing Adapt Tests

Angelyr:
- removed use of double negative
- fixed failing adapt test


mds/mds.c
```diff
@@ -579,7 +579,11 @@ static mds_id add_ent(struct mds* m, int t, mds_id* from)
 
 static void check_ent(struct mds* m, mds_id e)
 {
-#ifndef NDEBUG
+#ifdef NDEBUG // release build
+  /* good luck ! */
```
cwsmith: lol

PR #521, opened 2025-11-14
Update Sim Version Nightly Test

Angelyr:


cwsmith:
@Sichao25 Can you add this change to the local copy of the nightly script?

PR #534, opened 2026-04-20
support simModSuite 2026

cwsmith:
SimModSuite 2026.0-260411 replaces `GM_createFromAssemblyModel()`  with `ModelBuilder` APIs.  This PR makes those changes **and** moves and renames the header that provides the SimModSuite version to the top level dir of the repo (`apf_simConfig.h` -> `simModSuiteConfig.h`).  This is a breaking change... but I don't know if anyone will care unless they are using our header to determine which SimModSuite APIs to call etc.

A couple alternatives:
- keep `apf_sim/apf_simConfig.h.in` and have other packages depend on it... this seems messy as gmi_sim doesn't/shouldn't depend on apf_sim
- drop support for SimModSuite prior to 2023 (we support 15.0.191017 and later) and use the simmetrix `SimModSuiteVersion.h` header  - still a breaking change, but better in the long run to not rely on our own header for the same functionality

@bobpaw Thoughts?


bobpaw:
> SimModSuite 2026.0-260411 replaces `GM_createFromAssemblyModel()` with `ModelBuilder` APIs. This PR makes those changes **and** moves and renames the header that provides the SimModSuite version to the top level dir of the repo (`apf_simConfig.h` -> `simModSuiteConfig.h`). This is a breaking change... but I don't know if anyone will care unless they are using our header to determine which SimModSuite APIs to call etc.

Moving the file up sounds good to me. It can be a minor version bump if the `apf_simConfig.h` file is changed so that all it contains is `#include <simModSuiteConfig.h>`. Many compilers also support the `#warning` directive (standard since C23 [[1]](https://en.cppreference.com/c/preprocessor/error)) that can be used to issue a deprecation notice. I suggest that the name be prefixed like other PUMI headers since neither PUMI nor Simmetrix use header subdirectories.

> * keep `apf_sim/apf_simConfig.h.in` and have other packages depend on it... this seems messy as gmi_sim doesn't/shouldn't depend on apf_sim

I agree with the logic here.

> * drop support for SimModSuite prior to 2023 (we support 15.0.191017 and later) and use the simmetrix `SimModSuiteVersion.h` header  - still a breaking change, but better in the long run to not rely on our own header for the same functionality

The `simModSuiteConfig.h` could point to `SimModSuiteVersion.h` if it was found at configure time. It may also make sense to parse that file for version information instead of using the current install path heuristic. I did something similar a long time ago with the PDCurses library [[2]].

1. https://en.cppreference.com/c/preprocessor/error
2. https://github.com/bobpaw/FindPDCurses/blob/master/FindPDCurses.cmake

[2]: https://github.com/bobpaw/FindPDCurses/blob/master/FindPDCurses.cmake

apf_cap/pkg_tribits.cmake
```diff
@@ -2,8 +2,8 @@ tribits_package(SCORECapf_cap)
 
 include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 
-#directory containing apf_simConfig.h
-include_directories("${CMAKE_CURRENT_BINARY_DIR}")
+#directory containing simModSuiteConfig.h
+include_directories("${PROJECT_BINARY_DIR}")
```
bobpaw:
```suggestion
include_directories("$<BUILD_INTERFACE:${PROJECT_BINARY_DIR}>")
```

Correct me if I'm wrong, but I think this is only for the build tree and then should not be added to the include path for the install interface?

bobpaw:
Shouldn't this PR target develop?

bobpaw (approved):

cwsmith:
Thank you.
`pumi_` or `PUMI_` prefix for the header?  I'm leaning towards `pumi_`.

regarding use of `SimModSuiteVersion.h`:
To keep things simple, I'd rather just have one approach: the current extraction from the install path or directly using `SimModSuiteVersion.h`.

bobpaw:
Looks good. Will there be a tag/release or will you wait to bundle it with anything else?

cwsmith:
Good point.  This would be a minor version bump given that is is backward compatible. I'll push a commit to update the version but don't think we need to tag/release for this.

